Issue #2, opened 2022-03-06
Make vote locker upgradeable


tomlinton:
~~Immediately allow all withdrawals of locked up tokens, block new lockups, and set vote power to 0 (totalSupply and all balanceOf). The purpose of this is if we ever need to deprecate the contract.~~

We may never need that feature above, so lets just make the contract upgradeable so we can implement it if necessary.

tomlinton:
@sparrowDom I've changed this issue.

sparrowDom:
Thanks for the heads up 👍 done via: https://github.com/OriginProtocol/ousd-governance/pull/15

PR #15, opened 2022-03-14
Make VoteLocker upgradeable

sparrowDom:
This makes voteLocker upgradeable. 

**issue:** https://github.com/OriginProtocol/ousd-governance/issues/2


sparrowDom:
@tomlinton in OUSD we are using the `hardhat-deploy` extension that helps with execution of migration scripts. 

Also there is a `@openzeppelin/hardhat-upgrades` It would probably save us some time if we were to use that with OUSD. You need to deploy the first version of your Proxy contract using that module in order for it to function (can't migrate to it later). And it to some extent automates the Proxy contract deployment. Also runs the storage slot checker before each upgrade of a contract as a safeguard.

Do you have any opinions in using hardhat/JS for contract deployment instead of brownie? 

Seems like there are also [tools being developed](https://github.com/brownie-mix/upgrades-mix) for Brownie to use OpenZepplin upgrades but look as if they are still in their infancy.

tomlinton:
I was planning on just using brownie for the deployments. I don't think we'll need to deploy/upgrade very often or at all. OUSD might be a bit different as upgrades happen more often. I guess I could be convinced otherwise though 😁 Maybe @DanielVF also has thoughts.

tomlinton:
Made a bit of a mess of your branch sorry, mainly just added UUPSUpgradeable and some tests, but ended up committing some stuff by accident. Needed it in master anyway so will merge it.

Issue #9, opened 2022-03-07
Detect proxy contracts in proposal creation UI and allow use of implementation methods


tomlinton:
When creating a proposal some of the contracts listed are proxies. Generally, a user won't want to call the proxy methods, and would rather call the implementation methods. We should detect if the contract is a proxy and provide an option to select from implementation methods.

tomlinton:
Closed with thanks to @tomhirst in #18.

PR #18, opened 2022-03-15
New proposal proxy contract UI

tomhirst:
This PR addresses issue #9.

**I've added basic UI that**

1. Detects selection of a proxy contract from the ABI list (uses string matching right now) then picks out the associated implementation contract data

2. Informs the user that the current contract selected is a proxy and provides an option to select from the associated implementation contract's methods. If no associated implementation is available in the ABI list (IE `OGNStakingProxy`) an appropriate fallback message is displayed

3. On submission, lists/stores (localStorage) proposal actions alongside the relevant contract depending on which option (and method) the user selected

**Queries**

1. Do we want to detect proxy contract selection/availability on-chain? My understanding is that implementation ABIs can't be compiled from proxy contract address references alone. I guess we could do something through Etherscan should the implementation contract be verified. LMK if I'm thinking too far beyond the scope of this issue.

2. If accompanying implementation contract data isn't available in the ABI list, is a fallback message needed at all?

**Screenshots**

Screenshot 2022-03-15 at 18 39 34

Screenshot 2022-03-15 at 18 39 42

Screenshot 2022-03-15 at 18 39 49

Screenshot 2022-03-15 at 18 39 57

Screenshot 2022-03-15 at 18 40 11

Screenshot 2022-03-15 at 18 54 37


client/components/proposal/ProposalActionAddModal.tsx
```diff
@@ -36,6 +48,18 @@ export const ProposalActionAddModal = ({
         </ul>
         {step === 0 && (
           <AddActionContractForm
+            onChange={(data) => {
+              setIsProxy(data.name.includes('Proxy'));
```
tomlinton, line 52: I think if we check the contract abi for an `implementation()` method, call that, and search the contracts for an address matching the result of that call. If it doesn't find anything we should display an error (e.g. Contract is a proxy but no implementation found for <address>).

client/components/proposal/AddActionFunctionForm.tsx
```diff
@@ -90,6 +97,27 @@ export const AddActionFunctionForm = ({
 
   return (
     <form onSubmit={submitHandler}>
+      { isProxy && (
+        <>
+        { implementationAddress ? (
+          <div className="form-control w-full mt-2">
+            <span className="text-xs m-1 font-semibold">Contract selected is a proxy</span>
+            <label className="label bg-gray-100 p-3 rounded-lg cursor-pointer" htmlFor="implementation-methods">
+              <span className="label-text">Select from implementation methods</span>
```
tomlinton, line 106: Lets make this a dropdown (similar to the one on the first page) called `Implementation contract` that preselects the implementation that is detected.

Our vault uses a double proxy setup so the proxy implementation will be detected as the Vault contract, but that doesn't include a lot of callable methods from VaultCore. So I think we should allow them to select any contract for the implementation to handle that (unless you have a nicer way).

client/components/proposal/ProposalActionAddModal.tsx
tomhirst: This makes sense. Have introduced a mainnet call to grab the implementation address of all proxy contracts (those with an `implementation()` method), alongside loading UI and a relevant error message should no ABI matching the fetched address be present in our ABI list.

tomhirst:
@tomlinton Further improvements made as suggested. LMK if this is now what you were expecting.

**Screenshots**

If a proxy is selected but no associated ABI can be found, the user will get this error:

Screenshot 2022-03-16 at 14 42 08

If there is an available proxy UBI, post-loading (loader component and disabled submit button), no change to the UI will be seen:

Screenshot 2022-03-16 at 14 42 17

If a proxy is selected at stage 2, the default on the Implementation contract select box is the relevant implementation contract:

Screenshot 2022-03-16 at 14 42 23

With associated functions:

Screenshot 2022-03-16 at 14 42 28

However, the user has ultimate control to change the implementation contract (and thus the methods available) should they wish (even in stage 2):

Screenshot 2022-03-16 at 14 42 37

client/components/proposal/AddActionFunctionForm.tsx
tomlinton: Great. Yea the current vault setup doesn't implement `implementation` for both, and I haven't seen any other contracts built in the same way, so it is in itself an edge case unfortunately.

client/constants/index.ts
```diff
@@ -45,3 +45,7 @@ export const voteLockerContract = new ethers.Contract(
   ],
   provider
 );
+
+export const INFURA_ID = "460f40a260564ac4a4f4b3fffb032dad";
```
DanielVF, line 49: Is this okay to be here?

client/components/proposal/ProposalActionAddModal.tsx
```diff
@@ -36,9 +45,63 @@ export const ProposalActionAddModal = ({
...
+
+                  // Check that the implementation address exists in our contracts list
+                  const implementationData = contracts.find(
+                    ({ address }) => 
+                      address === implementationAddress
+                  );
+
+                  // If present, populate state
+                  if(implementationData) {
+                      setHasImplementationAbi(true);
```
DanielVF, line 84: nit: indentation

client/constants/index.ts
tomhirst: @DanielVF I left it hardcoded as that's how it was originally, just in another file (not as a constant).

We probably should move this out to an env var eventually. Should we raise an issue to introduce dotenv + an .env file?

client/components/proposal/ProposalActionAddModal.tsx
tomhirst: @DanielVF Fixed pushed.

Issue #16, opened 2022-03-14
Support delegation in VoteLockerCurve


tomlinton:

sparrowDom:
PR here: https://github.com/OriginProtocol/ousd-governance/pull/19

sparrowDom:
A better attempt at this issue in this PR: https://github.com/OriginProtocol/ousd-governance/pull/37

sparrowDom:
We've decided to hold off this one until we get the basic version of governance launched and then revisit.

tomlinton:
Closing this because we implemented the alternate token decay model.

PR #19, opened 2022-03-16
Support vote delegation

sparrowDom:
Adds delegation support: 
- users can delegate their votes to another address
- also works for previous block lookups of voting power
- currently max 100 delegators can delegate to 1 address
- adds function to cleanup delegators who have no voting power
- adds `console.sol` contract for debugging contract states

**Issue:** https://github.com/OriginProtocol/ousd-governance/issues/16

**Explanation why this can not be achieved only by Checkpoint modification:**

First idea was to only modify user Checkpoints when delegating. But that would complicate things to some extent since Checkpoint can only define one straight line. And with delegation voting power can become more complex than a single line. See graph below when summing together 2 voting power graphs: 

Lines

So to achieve the desired effect each user still controls their own checkpoint. And when calling `balanceOf` & `balanceOfAt` checks are made to correctly sum together appropriate user voting powers.


tomlinton:
This solution adds an unbounded loop into the balanceOf calculations. There is no constraints on the number of delegators, so I could cripple the governance by delegating from many accounts to a single user.

The checkpoint slopes can be adjusted by delegations in the same way that they are adjusted when a user increases the locked token amount of an existing lockup. It writes a new checkpoint which defines a new line.

contracts/VoteLockerCurve.sol
```diff
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: MIT
 
 // Inspired by the Curve.fi VotingEscrow contract and the mStable Solidity fork.
-//    - Adds compatability with the OpenZeppelin governance stack (4.5.0)
+//    - Adds compatibility with the OpenZeppelin governance stack (4.5.0)
 //    - Removes the separate lockup amount and lockup duration extension functions info
-//      favour of a single function for create or update.
```
tomlinton, line 6: Discrimination 💀

contracts/VoteLockerCurve.sol
```diff
@@ -15,8 +15,10 @@ pragma solidity ^0.8.2;
 
 import "OpenZeppelin/openzeppelin-contracts@4.5.0/contracts/token/ERC20/ERC20.sol";
 import "OpenZeppelin/openzeppelin-contracts@4.5.0/contracts/utils/math/SafeCast.sol";
+import "OpenZeppelin/openzeppelin-contracts@4.5.0/contracts/utils/cryptography/ECDSA.sol";
```
tomlinton, line 18: This important isn't need I don't think?

contracts/VoteLockerCurve.sol
```diff
@@ -50,6 +56,13 @@ contract VoteLockerCurve {
     uint256 public constant MAX_LOCK_TIME = 4 * 365 * 86400; // 4 years
     /// @notice Vote boost if locked for maximum duration
     uint256 public constant MAX_VOTE_MULTIPLE = 4;
+    /// @notice Maximum number of delegators per account
+    uint256 private constant MAX_DELEGATORS = 100;
```
tomlinton, line 60: We need to write tests for this. Although the views won't have an issue, the functions that call balanceOf and balanceOfAt that write state (e.g. the governance castVote is probably one) could become unusable if a full 100 delegators means that the function uses too much gas.

We need to try and figure out that most gassy function that calls the views that have the delegation loops, and write a test for the very worst case scenario. Even if the function is still callable we may need to reduce this number because we don't want voting for someone with a lot of delegations to cost a fortune.

contracts/VoteLockerCurve.sol
sparrowDom: yup thanks removing

contracts/VoteLockerCurve.sol
tomlinton: There is a whole bunch of these words that the US decided to change, just to be different. Join team British English though, we have better tea.

sparrowDom: Joining team British English then. For the tea and classiness

contracts/VoteLockerCurve.sol
sparrowDom: Thanks for pointing this out. After some testing I hate my implementation... So the bottom line: 
- each delegate adds ~50k gas to the `castVote` call
- can't test with more than 8 delegates using brownie+hardhat (even standalone node + brownie console in separate terminal) because it crashes hardhat. Brownie calls `debug_traceTransaction` [after tx execution](https://github.com/eth-brownie/brownie/blob/3aecd87f47c9c316c85b0b0c6252ff7d900cca74/brownie/network/transaction.py#L634) and hardhat [rolls over and dies](https://capture.dropbox.com/MJ4QCk5aEH9ArF9x) since the JSON it is trying to stringify is too large.

So I have reduces the max amount of delegators to 8. Obviously this approach has huge issues.

sparrowDom: I wonder if we should reduce all the checkpoints (voting graphs) of all delegators to one structure - similar to global checkpoints. But I fear that will put some heavy complexity to the contract code.

Need to think more about this one...

tomlinton:
We'll leave this out of the initial release and revisit later.

sparrowDom:
closing in **favour** of https://github.com/OriginProtocol/ousd-governance/pull/37

PR #23, opened 2022-03-16
Fix indentation

tomhirst:


tomhirst:
Closed as already fixed in master.

tomlinton:
Thanks for the fix @tomhirst. I added prettier and ran it over everything to take care of it. Also added GH checks so we'll get an error next time.

tomhirst:
Yep, I noticed after! Great work.

PR #25, opened 2022-03-29
Comments & cleanup

sparrowDom:
- removes some imports that are already imported by other contracts. And some functions that as a result of that don't need to be overriden
- adds comments to vote locker logic


tomlinton (approved):
Nice!

Issue #1, opened 2022-03-06
Add canceller role to timelock controller


tomlinton:
We want to extend the timelock to allow cancellation of in flight changes. This will involve adding the role and a method to cancel operations that have already been added to the timelock. The canceller will be a 2 of n multisig.

sparrowDom:
Ooooor just wait long enough and ppl from OpenZepplin will [do it themselves](https://github.com/OpenZeppelin/openzeppelin-contracts/commit/8b162e39b598924ef611dba90ff9d0aa0b676ecc)  

Seems by default that proposers get [also the canceller roles](https://github.com/OpenZeppelin/openzeppelin-contracts/commit/8b162e39b598924ef611dba90ff9d0aa0b676ecc#diff-8229f9027848871a1706845a5a84fa3e6591445cfac6e16cfb7d652e91e8d395R88-R90).  From what I can tell all that is needed is to revoke that role and add it to another account.

Checking their release schedule for the last 5 tags in Github they seem to do releases ~monthly. And last version (one we are using) has the last commit on Feb 9th. I would expect them to add the above code to a release in the next week or two. If ok by you @tomlinton I would wait till then and upgrade OZ contracts. Except if you have some reservations or would like to see this implemented quicker.

tomlinton:
Oh haha, that is great. Yea I propose we just wait. We have an audit beginning April 4th and we'll have to provide a commit some time before then, but worst case we just cherry pick this stuff into our own contracts and hope it doesn't change between then and the official OZ release.

tomlinton:
Marked blocked while we wait for that release.

sparrowDom:
No new release from OZ yet. They claim on the forum to have a [5 week release cycle](https://forum.openzeppelin.com/t/release-cycle/139). Seems they are slightly behind this time...

sparrowDom:
Pr for this: https://github.com/OriginProtocol/ousd-governance/pull/26

tomlinton:
Closed #26

PR #26, opened 2022-03-30
update OZ contracts

sparrowDom:
Updates OZ contracts to current master commit. Once new tag is released we can default to that one. 
- adds extra param to `_castVote` overridden function 

**issue:** https://github.com/OriginProtocol/ousd-governance/issues/1

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,10 +16,14 @@ pipx install eth-brownie
 ```
 
 ## Install brownie dependencies
+We require OpenZeppelin contract code that is in master and will be released after 4.5.0. Since the release
+tag has not been created yet we need to import them by commit hash. Once newer version of 4.5.0 is released
+we can default to that.
 
 ```bash
-brownie pm install OpenZeppelin/openzeppelin-contracts@4.5.0
-brownie pm install OpenZeppelin/openzeppelin-contracts-upgradeable@4.5.0
+
+brownie pm install OpenZeppelin/openzeppelin-contracts-upgradeable@a16f26a063cd018c4c986832c3df332a131f53b9
+brownie pm install OpenZeppelin/openzeppelin-contracts@02fcc75bb7f35376c22def91b0fb9bc7a50b9458
 ```
 
 ## Install hardhat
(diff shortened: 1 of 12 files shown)


tomlinton:
Nice, any idea why the tests fails? Might need to modify `brownie-config.yaml`?

Also would be great if we could add a test for the timelock canceller.

README.md
```diff
@@ -16,10 +16,14 @@ pipx install eth-brownie
...
 ## Install brownie dependencies
+We require OpenZeppelin contract code that is in master and will be released after 4.5.0. Since the release
+tag has not been created yet we need to import them by commit hash. Once newer version of 4.5.0 is released
+we can default to that.
 
 ```bash
-brownie pm install OpenZeppelin/openzeppelin-contracts@4.5.0
-brownie pm install OpenZeppelin/openzeppelin-contracts-upgradeable@4.5.0
+
+brownie pm install OpenZeppelin/openzeppelin-contracts-upgradeable@a16f26a063cd018c4c986832c3df332a131f53b9
```
tomlinton, line 25: What's the difference between these two commit hashes? Is one 4.5.0 and one the hash with timelock canceller included? Could we have left one of these as 4.5.0?

sparrowDom: They are 2 separate github repos. Have just taken latest commit hash from each of the repos

sparrowDom:
have added a test for this. Should be in good condition to merge now @tomlinton . Sorry for taking so long to get this done, got distracted understanding how all the governance contracts/modules interconnect.

README.md
tomlinton: I'm a bit torn here, on the way hand we probably don't need anything from master for the upgrade contracts that isn't in 4.5.0, on the other hand, the current master is probably closer to to the next release than 4.5.0. I think we'll probably stick with this.

tomlinton:
Great job @sparrowDom thank you!

tomlinton (approved):

README.md
sparrowDom: Yeah I know it is not an ideal solution. Still looking forward, it will be probably easier for us to transition to 4.6.0 using current master from OZ, than it would be if we were to implement canceller roles ourselves. 

Other way would be cherry picking only the canceller role pull request. Haven't tried that, could be there would be some merge conflicts.

PR #36, opened 2022-04-04
Vote power grouping refactoring

sparrowDom:
Intention of this PR is to prepare grounds for second attempt at voting power delegation. Grouping multiple votes together is separated into a separate contract. And all the data required for that is combined together in a `GroupVotePowerState` struct. 

The ideas is that `GroupVotePowerState` can also be used for delegation where multiple delegator votes to a same delegatee account can be grouped together in a similar manner as total votes of a contract are grouped/reduced together.


tomlinton:
Sorry was a bit slow on this @sparrowDom, could you add a bit of context on what a group is and how it works? How does grouping handle the linear decay being different for different epochs?

sparrowDom:
Hey @tomlinton, the group is extracted piece of logic that in the current master takes care of supporting `totalSupply` and `totalSupplyAt` functions. It is basically a way to mathematically represent a combination of many vote decaying functions reduced into one structure. 

This image shows it best(With the risk of looking like a broken record, reusing this image too many times):
158651631-7a6ce8e1-2d0c-477b-80aa-d97ef592fef0

This is what `totalSupply` is represented like in the current master. I have just extracted it into a separate contract and named it groupVoteLocker. Separating it this way makes it easy to use it for other purposes where multiple decaying voting power functions need to be reduced into 1 structure that is more gas effective. Which turns out to be ideal for voting delegation.

We can have a 1:1 VC to go into more detail if you'd like?

sparrowDom:
ALso this PR builds on top of the first one: 
https://github.com/OriginProtocol/ousd-governance/pull/37

tomlinton:
Got it, thanks!

sparrowDom:
closing since all of the code in this PR is included in https://github.com/OriginProtocol/ousd-governance/pull/37

PR #37, opened 2022-04-05
Second take on delegation

sparrowDom:
This delegation implementation makes delegation gas usage much more scalable. Build off of: https://github.com/OriginProtocol/ousd-governance/pull/36

Open issues with this PR: 
- test vote power expected tolerances needed to be increased to 2%. 

**Gas cost increases:**
Without the delegation changes (current master) the `lockup` function costs between `315k-350k`. With delegation enabled (this PR) the `lockup` function costs between `350k - 505k`. The `delegate` function costs `216k` gas.


contracts/VoteLockerCurve.sol
```diff
@@ -103,38 +100,19 @@ contract VoteLockerCurve is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
     ///@notice Lockup mapping for each user
     mapping(address => Lockup) public lockups;
 
-    /* @notice slopeChanges part of the equation to define combined voting power of
-     * all users. Slope changes always complement only the latest global Checkpoint and are
-     * not used when fetching combined voting power of previous blocks.
-     */    
-    mapping(uint256 => int128) public slopeChanges;
+    ///@notice Contains group vote information used for totalSupply & totalSupplyAt
+    GroupVotePowerState public totalVotes;
```
franckc, line 107: What is the approximate gas cost for a user lockup action now that we have to update both the `totalVotes` and `userTotalVotes` structures?

sparrowDom: good point thanks for bringing this up, have updated that info in the original PR description

contracts/VoteLockerCurve.sol
```diff
@@ -318,14 +289,35 @@ contract VoteLockerCurve is Initializable, OwnableUpgradeable, UUPSUpgradeable {
...
+        // take away voting power from current delegatee or user itself
+        _writeGroupCheckpoint(
+            -lastCheckpoint.slope,
+            -lastCheckpoint.bias,
+            userTotalVotesWithDelegation[currentDelegatee != address(0) ? currentDelegatee : msg.sender]
+        );
+        // add voting power to new delegatee or user itself
+        _writeGroupCheckpoint(
+            lastCheckpoint.slope,
+            lastCheckpoint.bias,
```
sparrowDom, line 318: same thing applies here

client/.gitignore
```diff
@@ -35,3 +35,6 @@ yarn-error.log*
 
 # typescript
 *.tsbuildinfo
+
+# localhost artifacts
+networks/*.localhost.json
```
tomlinton, line 40: Ty 🤣

tomlinton:
This is a really cool way to do the delegation, nice work @sparrowDom. As you noted it does reduce the accuracy of the user balances, and we'll need to think a bit deeper about the potential consequences and whether there are certain scenarios that this will cause problems under.

contracts/BaseVoteLocker.sol
```diff
@@ -0,0 +1,38 @@
...
+     */
+    function max(uint256 _a, uint256 _b) internal pure returns (uint256) {
+        return _a >= _b ? _a : _b;
+    }
+
+    /**
+     * @dev Returns the smallest of two numbers.
+     * @param _a First number
+     * @param _b Second number
+     * @return Smallest of _a and _b
```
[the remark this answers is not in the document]
sparrowDom: haha yeah tnx

contracts/BaseVoteLocker.sol
```diff
@@ -0,0 +1,38 @@
...
+     * @param _a First number
+     * @param _b Second number
+     * @return Largest of _a and _b
+     */
+    function max(uint256 _a, uint256 _b) internal pure returns (uint256) {
+        return _a >= _b ? _a : _b;
+    }
+
+    /**
+     * @dev Returns the smallest of two numbers.
```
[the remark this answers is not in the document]
sparrowDom: got it thanks

contracts/GroupVoteLocker.sol
```diff
@@ -0,0 +1,331 @@
...
+import "OpenZeppelin/openzeppelin-contracts@02fcc75bb7f35376c22def91b0fb9bc7a50b9458/contracts/utils/math/SafeCast.sol";
+import "OpenZeppelin/openzeppelin-contracts@02fcc75bb7f35376c22def91b0fb9bc7a50b9458/contracts/utils/Strings.sol";
+import "./console.sol";
+
+contract GroupVoteLocker is BaseVoteLocker {
+    ///@notice Checkpoint structure representing linear voting power decay
+    struct Checkpoint {
+        int128 bias;
+        int128 slope;
+        uint256 ts;
```
DanielVF, line 15: 64 bit timestamps are already good until the sun explodes. If we switch this to 128 bit, and the block number to 128 bit, then we will save a storage slot here in the struct that is going to have tons of reads and writes.

contracts/GroupVoteLocker.sol
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    /**
+     * @dev Binary search (bisection) to find epoch closest to block.
+     * @param _block Find the most recent point history before this block
+     * @param _maxEpoch Maximum epoch
+     * @return uint256 The most recent epoch before the block
+     */
+    function _findEpoch(
+        Checkpoint[] memory _checkpoints,
```
DanielVF, line 314: _findEpoch is designed to efficiently search inside a list of checkpoints. However, rather than searching storage slots, it works from memory. This means that before it can begin searching, it first *loads every single checkpoint struct from storage*.

Not only does it have to read every storage slot before being called, it's reading all the fields of these structs in, even the ones that won't be looked at.

This way of working probably needs a change.

We need to make sure we have unit tests that work on 400 checkpoint systems.

contracts/GroupVoteLocker.sol
```diff
@@ -0,0 +1,331 @@
...
+        internal
+        view
+        returns (uint256)
+    {
+        Checkpoint memory lastCheckpoint = _checkpoint;
+
+        // Floor the timestamp to weekly interval
+        uint256 iterativeTime = _floorToWeek(lastCheckpoint.ts);
+        // Iterate through all weeks between _checkpoint & _time to account for slope changes
+        for (uint256 i = 0; i < 255; i++) {
```
DanielVF, line 277: I think you noted this elsewhere, but I don't think we can do a loop here, since this is going to go wrong after five years, (and it's not very gas efficient.

Can we just multiply everything that's not the final section? (Making sure rounding works with the rest of the system)

DanielVF, line 277: So this will break everything if we go for five years without a checkpoint? 

Do have we add a way to force adding intermediate checkpoints from some method?

DanielVF: Can we skip this loop if instead of a mapping to group slope changes, we somehow work of the checkpoints that have both a bias and a slope?

contracts/GroupVoteLocker.sol
```diff
@@ -0,0 +1,331 @@
...
+     */
+    function _findEpoch(
+        Checkpoint[] memory _checkpoints,
+        uint256 _block,
+        uint256 _maxEpoch
+    ) internal view returns (uint256) {
+        uint256 minEpoch = 0;
+        uint256 maxEpoch = _maxEpoch;
+        for (uint256 i = 0; i < 128; i++) {
+            if (minEpoch >= maxEpoch) break;
```
DanielVF, line 321: As far as I can tell, this searches "from the bottom up". The most expensive gas search would be for search for the most recent checkpoint. 

If this were used for searching future checkpoints, this would be efficient, but it looks like when it's used, it required that it is searching older blocks.

Sill correct, may or may not make much difference, just something I noticed.

contracts/GroupVoteLocker.sol
```diff
@@ -0,0 +1,331 @@
...
+            groupState.epoch
+        );
+
+        if (recentGroupStateEpoch == 0) {
+            return 0;
+        }
+
+        Checkpoint memory checkpoint0 = groupState.checkpoints[recentGroupStateEpoch];
+
+        if (checkpoint0.blk > _blockNumber) {
```
DanielVF, line 224: Should this be a require? _findEpoch is supposed to always return something before the timestamp, unless it's the 0 epoch, and we have check above for the zero epoch. Theoretically, hitting this conditional should be impossible?

contracts/GroupVoteLocker.sol
```diff
@@ -0,0 +1,331 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.4;
+
+import "./BaseVoteLocker.sol";
+import "OpenZeppelin/openzeppelin-contracts@02fcc75bb7f35376c22def91b0fb9bc7a50b9458/contracts/utils/math/SafeCast.sol";
+import "OpenZeppelin/openzeppelin-contracts@02fcc75bb7f35376c22def91b0fb9bc7a50b9458/contracts/utils/Strings.sol";
+import "./console.sol";
+
+contract GroupVoteLocker is BaseVoteLocker {
```
DanielVF, line 10: There are a lot of `public` methods in this that can probably be `external`. Should make the contract deploy size smaller.

contracts/GroupVoteLocker.sol
tomlinton: > 🟥 Given that anyone can create as many global snapshots as they want per block, and this has a linear number of storage reads per snapshot, this could be a contract killer here.

The loop is bounded so I think in the worst case you'll end up with it finding the wrong epoch rather than a broken contract. The consequences of that are that the historical total supply will break.

Exponential increases in checkpoint values only result in linear increases in gas cost here, so I think it's probably unfeasible to use this as an attack vector. We should write some tests though.

tomlinton:
Great feedback @DanielVF, just want to note that most of these applies to the VoteLockerCurve contract as it is largely the same logic.

sparrowDom:
Thanks for great and thorough feedback @DanielVF 👍

contracts/GroupVoteLocker.sol
```diff
@@ -0,0 +1,331 @@
...
+    struct Checkpoint {
+        int128 bias;
+        int128 slope;
+        uint256 ts;
+        uint256 blk;
+    }
+
+    ///@notice Stores token lockup details
+    struct Lockup {
+        int128 amount;
```
DanielVF, line 21: Why is Lockup.amount an int128? Can people do negative lockups?

tomlinton: This was taken straight from the Curve contract. I didn't attempt to change the structure.

contracts/GroupVoteLocker.sol
DanielVF: @tomlinton The problem isn't the loop - the problem is the `memory` attribute here. This will load EVERY element in from storage before running the function. It's an implicit loop over every checkpoint ever stored in this array.

tomlinton: Ahh I see what you mean 👍

sparrowDom:
Shelving this for now due to radioactive token implementation. Lets re-open it if need be.

contracts/GroupVoteLocker.sol
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+
+    /**
+     * @dev Add a checkpoint to the group state
+     * @param userSlopeDelta Change in slope that triggered this checkpoint
+     * @param userBiasDelta Change in bias that triggered this checkpoint
+     * @param groupState group state to have a checkpoint added
+     */
+    function _writeGroupCheckpoint(
```
DanielVF, line 99: 🔴 I've not thoroughly checked out to see if it's a bug, but it feels like it might be one. Since we always write at least one checkpoint when write is called, there can be more than one checkpoint for a single block.

When searching for a block's checkpoints, there is a deterministic algorithm that runs to find the checkpoint, however it does not guarantee that it will select the last checkpoint per block, and could select any of the block's checkpoints. Almost everything we do involves comparing values at a time of two different checkpoint lists against each other. An attacker could possibly manipulate this such that the the user and the total values for a block were from different checkpoint orders, and thus bad things could happen.

contracts/GroupVoteLocker.sol
sparrowDom: I think this uses bisection for searching

contracts/GroupVoteLocker.sol
sparrowDom: Definitely something to investigate. Though I think `epochs` are used rather than solely block numbers to find checkpoints. And each time a new one is created an epoch increases as well.

PR #48, opened 2022-04-13
Sparrow dom/max loss

sparrowDom:
- add max loss to the governance proposal
- run the node in forked environment, so that all the contracts from the mainnet are available through node
- make new proposal page render if opened directly
- fix a problem where previously it wasn't possible to submit a reallocation proposal


tomlinton (approved):
Thank you sir!

PR #49, opened 2022-04-19
Add accurate useEffect deps to fix infinite loop on proposals index

tomhirst:
With an empty set of proposals locally, the proposals index template crashes the app. This hotfix adds accurate dependencies to prevent infinite re-rendering, allowing the page to load.


tomlinton (approved):
Thanks @tomhirst!

PR #51, opened 2022-04-21
in development mode reset listener to current chain block when necessary

sparrowDom:
in development mode reset listener to current chain block when necessary


sparrowDom:
@tomlinton merging this in for making it more convenient for development in forked hardhat mode. Will address any issues in later PRs.

PR #55, opened 2022-04-22
Add blank governance.localhost.json

tomlinton:
We need to keep this file around or the deploy breaks.


tomlinton:
Yea it's an import that the production build breaks on if non existent. There might be some conditional import thing you could do or otherwise handle it in the production build to avoid us needing to include it, but the easiest option for now is just a blank one.

Issue #61, opened 2022-04-25
Show lock end date


joshfraser:
I created a lockup for 1 week and the lockup shows as ending in 0 weeks. We should probably show the date here instead, or show a specific date in addition to the week counter.

![](https://cdn.discordapp.com/attachments/758123420230615080/968197496524382228/Screen_Shot_2022-04-25_at_1.02.35_PM.png)

tomhirst:
Closing now #65 is merged.

PR #65, opened 2022-04-26
Add lockup end date alongside weeks count

tomhirst:
Addresses #61

Replacing the large week counter figure with a date string didn't look right, so I added the end date underneath in addition. This approach allows the alignment of the figures and their unit types to remain uniform, too.

This will show for any existing lockup (not just those of < 1 week) as it's useful to know the exact date no matter how many weeks your tokens are locked up for.

Screenshot 2022-04-26 at 12 16 31


joshfraser:
Hmm, I don't love how this looks. What if the large text said "0 weeks" and we replaced the "Weeks" label with the date?

Nit, but let's add a comma between the date and the year.

tomhirst:
> Hmm, I don't love how this looks. What if the large text said "0 weeks" and we replaced the "Weeks" label with the date?
> 
> Nit, but let's add a comma between the date and the year.

Makes sense! Implemented:

Screenshot 2022-04-27 at 12 17 38

tomlinton (approved):

PR #66, opened 2022-04-26
Token amount formatting

tomhirst:
This introduces uniformity when formatting at a glance token amounts.

Instead of the long decimal numbers on the Vote Escrow and Proposal pages, I've truncated them to be easier on the eye and prevent the overflow problem we were experiencing.

Screenshot 2022-04-26 at 12 51 49


tomlinton (approved):
Thanks for cleaning up my mess!

Issue #63, opened 2022-04-25
Missing error handling on number of weeks


joshfraser:
The max weeks you can lock up for is 208, but you'd never guess that from the UI feedback.

<img width="674" alt="Screen Shot 2022-04-25 at 1 03 11 PM" src="https://user-images.githubusercontent.com/982234/165140854-e288a1a6-1f45-487f-967c-1ec50d7e0220.png">

It might be better for us to use an input slider here, instead of an open text input.

tomlinton:
Yea I like the slider idea. The UI around the length is a bit nasty, I thought it needed a bit more input from someone who is better at it than me. Some options I considered:

- Lockup length specified in days with slider
- Lockup length in months or quarterly or bianually that snaps to each

I like something like a biannual interval to normalise longer lockups in the users mind.

tomhirst:
> Months seems like a reasonable increment to me. My brain instantly knows how long 48 months is, but I need to stop and think for a second to process 208 weeks.
> 
> The slider should default to 48 months of course.
> 
> We should also add some information to this page about the % of OGV that is locked up and the average lock time. We expect both these numbers to be high, which should encourage others to lock up as well.
> 
> ve(3,3)

All this makes sense and I agree, it's easier to grok the length of time in months.

Should we merge the UI part of this issue first (https://github.com/OriginProtocol/ousd-governance/pull/67), then we can open 2 new issues:

1. Swap out the weeks component logic in favour of months
2. Add an info panel detailing the current token lockup stats

tomlinton:
Agreed, I think #67 is a good start and we can refine it.

PR #67, opened 2022-04-26
Lockup UI

tomhirst:
Addresses #63 

This implements a range input to prevent the max. amount of weeks allowed to lockup tokens being surpassed. The markers underneath represent biannual lockup periods (as suggested by @tomlinton).

The input is currently freeform, though we can set `step=26` if we want to enforce specific lockup periods from the UI.

Screenshot 2022-04-26 at 14 13 53


joshfraser:
This looks nice! Can we add some labels on the tick marks?  

1wk on the left, 4yr on the right and 1yr/2yr/3yr in between?

tomhirst:
> This looks nice! Can we add some labels on the tick marks?
> 
> 1wk on the left, 4yr on the right and 1yr/2yr/3yr in between?

Have added this in as below. Looks better!

I've labelled the far left marker as 0 wks to suit the current implementation. We can look at refactoring the logic if we prefer 1 week as a minimum value/default state in another PR.

Screenshot 2022-04-27 at 11 47 11

joshfraser:
0 weeks works well for the initial claiming (we'll reuse this component) since dragging it to zero will allow you to claim your OGV without adding any lock up.

Other than that scenario, we'll want to limit the slider to stop at 1 week minimum, but it's fine to show 0 weeks on the tick marks.

We should probably rethink how it works if you have an existing lockup that you are extending. Perhaps the slider shows your current lockup length? Or we can always default back to max time? I'm not sure, but we should think through this UX a bit.

tomlinton (approved):

Issue #62, opened 2022-04-25
Improve the UI for extending lockups


joshfraser:
it's unclear what will happen when i already have locked tokens & try to create a new lockup

<img width="691" alt="Screen Shot 2022-04-25 at 1 04 33 PM" src="https://user-images.githubusercontent.com/982234/165140197-2a3a322c-1c1d-4a28-9489-0b9dc4fdab91.png">

We should follow Curve's lead and make it more obvious that you have two options here:

 - extend the length of your existing lockup
 - add more tokens to your existing lockup.
 
<img width="564" alt="Screen Shot 2022-04-25 at 1 18 42 PM" src="https://user-images.githubusercontent.com/982234/165140412-f78adfde-4fb4-47a2-a3d7-5f36a87abfda.png">

people who are familiar with Convex's locking system will be particularly confused because Convex allows you to create multiple discrete lockups and we don't

PR #73, opened 2022-04-27
Improve UI for users with existing lockups

tomhirst:
Work to address #62 

Screenshot 2022-04-27 at 17 07 38


sparrowDom:
I like the simplicity of this approach. There are some things with step 3. I am not sure how they function. E.g.:
- user has locked up 100 OGV for 4 years. That happened 1 year ago
- today user has 10 OGV in wallet and wants to extend the lockup by 1 year and add 10 OGV to the existing lockup. (totalling 4 years from today)


Since user's balance is only 10 OGV we wouldn't be able to pre-populate the slider to pre-existing 100 OGV. 
User is already 1 year into their lockup. How do we pre-populate the slider? Set it to 4 years as it was done in the previous lockup. Or maybe to 3 years since 1 year has already passed.

tomhirst:
> 

Thanks, @sparrowDom! Good points.

The lockup amount slider max value populates by adding together the non-locked up OGV and lockup OGV of the user. The pre-populated current value is then set using the existing amount the user has locked up. They can extend up to their total balance, but not below what is already locked up. So I think we are OK there. Ref: https://github.com/OriginProtocol/ousd-governance/blob/5ec8a4349bd727f80efb44f13402f880a6c08292/client/pages/vote-escrow.tsx#L209 and https://github.com/OriginProtocol/ousd-governance/blob/5ec8a4349bd727f80efb44f13402f880a6c08292/client/pages/vote-escrow.tsx#L38

If a user is already 1 year (52 weeks) into their lockup, we pre-populate with now - 52 weeks. So if the lockup was for 2 years initially (104 weeks), and they were 1 year in (52 weeks), the lockup length slider would pre-populate with 52 weeks. They can then extend up to the maximum lockup length of 4 years (208 weeks). I think that makes us OK there too. Ref: https://github.com/OriginProtocol/ousd-governance/blob/5ec8a4349bd727f80efb44f13402f880a6c08292/client/pages/vote-escrow.tsx#L42

LMK if I'm missing something, have been staring at this for a while 😄

tomlinton:
> LMK if I'm missing something, have been staring at this for a while

I think you've covered everything here. The only thing I might be inclined to add is a bit of text that explains an existing lockup, e.g. "You have an existing lockup. You can increase the amount, the length, or both." That might not be necessary. It isn't immediately clear to me that the slider preset is my existing lockup, particularly because one says 401m and the preset is the full number , i.e. 400765...

I'm loving the UI around extending one or the other or both though, nice work 🎉

tomhirst:
> > LMK if I'm missing something, have been staring at this for a while
> 
> I think you've covered everything here. The only thing I might be inclined to add is a bit of text that explains an existing lockup, e.g. "You have an existing lockup. You can increase the amount, the length, or both." That might not be necessary. It isn't immediately clear to me that the slider preset is my existing lockup, particularly because one says 401m and the preset is the full number , i.e. 400765...
> 
> I'm loving the UI around extending one or the other or both though, nice work 🎉

Thanks for looking @tomlinton. You're right that it could be more obvious that the slider is preset to the existing lockup amount. I've added a small text note as you suggested, and, have given the slider the same numerical formatting as the top boxes.

My initial thoughts here were to show the full number to allow more granular input control, though, I think it's more important to have the numbers match in style. We are locked to summarizing in the top boxes because of space restraints, so I feel this is the best way forward.

I've committed some tidying up and a small fix that meant existing lockups would fail the smart contract validation when only the amount was changed (104 weeks instantly becomes 103 weeks as soon as it's set, so we increment the weeks by 1 for existing lockups to get the transaction through).

This is ready for pre-merge review now pending any further comments from the team.

tomlinton:
This looks great to me!

tomlinton (approved):

Issue #76, opened 2022-04-28
Vote Escrow: Add an info panel detailing the current token lockup stats


tomhirst:
We should also add some information to this page about the % of OGV that is locked up and the average lock time. We expect both these numbers to be high, which should encourage others to lock up as well.

Ref: #63

PR #78, opened 2022-05-04
Add token stats panel

tomhirst:
Addresses #76.

Adds a lockups table to the database and listens for Ethereum events to populate lockup data when added/amended. This lets us calculate the total supply of OGV locked up and the average lockup length in weeks.

Includes front-end UI as shown in the screenshot below:

Screenshot 2022-05-04 at 14 58 20

### DIFF
--- a/client/listener.ts
+++ b/client/listener.ts
@@ -42,7 +42,7 @@ const contracts = [
     name: "VoteLockerCurve",
     address: GovernanceContracts.VoteLockerCurve.address,
     abi: GovernanceContracts.VoteLockerCurve.abi,
-    events: ["LockupCreated"],
+    events: ["LockupCreated", "LockupUpdated"],
   },
 ];
 
@@ -62,6 +62,7 @@ const ethereumEvents = new EthereumEvents(web3, contracts, options);
 
 const handleEvents = async (blockNumber, events, done) => {
   for (const event of events) {
+    const now = event.timestamp;
     if (event.name == "ProposalCreated") {
       try {
         await prisma.proposal.create({
@@ -73,6 +74,36 @@ const handleEvents = async (blockNumber, events, done) => {
         logger.info("Inserted new proposal");
       } catch (e) {
         // Likely duplicate
+      }
+    } else if(event.name === "LockupCreated") {
+      try {
+        await prisma.lockup.create({
+          data: {
+            address: event.values.provider,
+            amount: event.values.amount,
+            end: event.values.end,
+            weeks: parseInt(((event.values.end - now) / 604800).toFixed()),
+          },
+        });
+        logger.info("Inserted new lockup");
+      } catch (e) {
+        // Likely duplicate
+      }
+    } else if(event.name === "LockupUpdated") {
+      try {
+        await prisma.lockup.update({
+          where: {
+            address: event.values.provider,
+          },
+          data: {
+            amount: event.values.amount,
+            end: event.values.end,
+            weeks: parseInt(((event.values.end - now) / 604800).toFixed()),
+          }
+        });
+        logger.info("Updated existing lockup");
+      } catch (e) {
+        // Likely not found
       }
     } else {
       try {
--- a/client/listener.ts
+++ b/client/listener.ts
@@ -91,17 +91,24 @@ const handleEvents = async (blockNumber, events, done) => {
       }
     } else if(event.name === "LockupUpdated") {
       try {
-        await prisma.lockup.update({
+        await prisma.lockup.updateMany({
           where: {
             address: event.values.provider,
           },
           data: {
+            active: false,
+          }
+        });
+        logger.info("Updated existing lockup(s)");
+        await prisma.lockup.create({
+          data: {
+            address: event.values.provider,
             amount: event.values.amount,
             end: event.values.end,
             weeks: Math.round((event.values.end - now) / 604800),
-          }
+          },
         });
-        logger.info("Updated existing lockup");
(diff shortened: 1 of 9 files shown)


client/listener.ts
```diff
@@ -74,6 +75,36 @@ const handleEvents = async (blockNumber, events, done) => {
...
         // Likely duplicate
       }
+    } else if(event.name === "LockupCreated") {
+      try {
+        await prisma.lockup.create({
+          data: {
+            address: event.values.provider,
+            amount: event.values.amount,
+            end: event.values.end,
+            weeks: parseInt(((event.values.end - now) / 604800).toFixed()),
```
sparrowDom, line 85: ParseInt is basically doing a Math.floor right? In case the transaction is mined a bit later could this potentially result in a week being just under a natural number?

Meaning that this code: `(event.values.end - now) / 604800` could result in a number like `5.999` and would be rounded down to 5. Or will `event.timestamp` alway correlate to a correct `event.values.end`?

So this is nitpicking a bit, but might be more in line with defensive programming to replace `parseInt` with `Math.round`

client/listener.ts
```diff
@@ -74,6 +75,36 @@ const handleEvents = async (blockNumber, events, done) => {
...
+    } else if(event.name === "LockupUpdated") {
+      try {
+        await prisma.lockup.update({
+          where: {
+            address: event.values.provider,
+          },
+          data: {
+            amount: event.values.amount,
+            end: event.values.end,
+            weeks: parseInt(((event.values.end - now) / 604800).toFixed()),
```
sparrowDom, line 101: same logic from above comment applies here

client/prisma/schema.prisma
```diff
@@ -34,3 +34,13 @@ model Listener {
 
     @@map("listener")
 }
+
+model Lockup {
+  id       Int @default(autoincrement()) @id
+  address  String @unique
```
sparrowDom, line 40: There might be more than 1 lockup per account

client/prisma/schema.prisma
```diff
@@ -34,3 +34,13 @@ model Listener {
...
+  amount   String
+  end      String
+  weeks    Int @default(0)
```
sparrowDom, line 43: There should always be a `weeks` value present when creating a Lockup object I think? If I am not mistaking I would remove the default value and rather have it fail with an error when no `weeks` present at object creation.

sparrowDom (commented):
Great job and super clean code 💪 I am looking forward to learning from you how to organise frontend code more neatly

Left a couple of minor comments below.

client/prisma/schema.prisma
tomlinton: Yea for some additional context here in case you weren't aware @tomhirst, @DanielVF is in the process of reworking on the lockup contract and it should support multiple lockups of different lengths. Not merged yet though of course, so I think we can merge this because it is currently correct and do the changes required for multiple lockups later as there'll be more than just this PR that needs updating.

sparrowDom: Another thought: currently you can still modify your existing lockup. It might make sense to add another boolean to the table e.g. `active` or `latest` that would indicate which is the last lockup user has created. This would enable us to (instead of modify existing lockup in the table to) have data for historical lockups on hand in case we would ever want to do any analysis based on that.

client/listener.ts
tomhirst: Good catch and makes sense as it's safer to derive a whole number with `Math.round`.

There's a bigger point here, too:

The final average weeks figure presented will rarely be a whole number (`total weeks / number of lockups`), so we could take the decimal on insert (as opposed to rounding at all).

Though, from a UX perspective, 46.57 weeks (or whatever) is always tougher to grok than 46 weeks or 46 weeks, 4 days.

I vote we go with rounded for now as that's how the inputs present lockup lengths. So I've implemented your suggestion in the listener and the front-end component that displays the average figure.

https://github.com/OriginProtocol/ousd-governance/blob/5cfbb423a31f1e42e138291659eeac8ceb633edf/client/components/vote-escrow/LockupStats.tsx#L28

client/prisma/schema.prisma
tomhirst: Agreed, updated.

client/prisma/schema.prisma
tomhirst: > Yea for some additional context here in case you weren't aware @tomhirst, @DanielVF is in the process of reworking on the lockup contract and it should support multiple lockups of different lengths. Not merged yet though of course, so I think we can merge this because it is currently correct and do the changes required for multiple lockups later as there'll be more than just this PR that needs updating.

Thanks, @tomlinton, got it. Yeah, we can merge this and then open a new issue to refactor the UI to match the contract changes. I can take ownership of that one when @DanielVF's work is merged.

> Another thought: currently you can still modify your existing lockup. It might make sense to add another boolean to the table e.g. `active` or `latest` that would indicate which is the last lockup user has created. This would enable us to (instead of modify existing lockup in the table to) have data for historical lockups on hand in case we would ever want to do any analysis based on that.

Smart, I like it. Would be nice to build a dashboard for users displaying this historic data.

Have extended this up as you suggested.

client/listener.ts
sparrowDom: Agree with your suggestion

client/prisma/schema.prisma
sparrowDom: sweet thanks 👍

sparrowDom (approved):
great job LGTM

PR #81, opened 2022-05-06
Draft: OGV Staking

DanielVF:


DanielVF:
## Half complete OgvStaking review


#### Requirements

The OgvStaking contract provides votes and rewards to stakers. The actual implimentation of the voting history is handeled by an OpenZepplin Library, and the rewards come from an external inflation source. Thus the primary concern of this contract's code is the staking, unstaking, and distributing rewards.

The rewards distrobution follows the Masterchef style, where an internal rewards per share accumulator increases rewards are given out.

The staking is done by using the end of the staking (start + duration - epoch) on an exponential curve. 

Key risks:
- the power function used in the staking
- the rewards distribution math.
- Both the rewards and the staking are in the same currency, which allows math errors to pull from the wrong pot.


### Deployment Considerations

- Will probably want some kind of monitoring / analytics.
- Tom handling deployement?
- Upgradable?

#### Internal Stored State

`epoch` is only set when creating the contract. Is is only read when calcuting the points to be added to a user, once to possibly move the start date forward, and once to see how long the end of the staking is from the epoch.

`ogv` is set when creating the contract. It is used to move OGV to be staked into the contract, and to return in when unstaked, and to send rewards to users that have already been sent to the staking contract.

`rewardsSource` is only set when creating the contract. It is used to preview and request inflation and other rewards to be sent to the staking contract.

`accRewardPerShare` is only set when rewards arrive - increased by the incoming funds / total supply. 

`rewardDebt[user]` must be calculated and writen at the end of every externaly callable function that changes a user's balance or gives rewards. It is read when determaning

`lockups[user][]` are pushed in staking, modified when extending, and cleared when unstaking. They are only read when extended or unstaked.


_(Remainder not yet uncomplete)_

- What can be always said about relationships between stored state
- What must hold true about state before a function can run correctly (preconditions)
- What must hold true about the return or any changes to state after a function has run.

- accRewardPerShare * totalSupply should always be greater than every users reward debt?
- rewardDebit should be zero after all unstakes for a user

Does this code do that?


#### Attack

The staking / unstaking transfers themselves seem fairly straightforward. The concern would be something that would cause an erronously large number of points to be returned for a stake.

I think the biggest risk in the rewards calculations. We may want to move holding the rewards to a seperate contract?


#### Logic

Emergency withdraws turning on noRewards. This looks correct, in that the rewards are skipped, and the rewardDebt will still be adjusted to match the current situation.

Staking to youself - currently untested


#### Tests

  - [x] Each publicly callable method has a test
  - [x] Each business if branch has a test
  - [ ] Each logical branch has a test
  - [ ] Each require() has a test
  - [ ] Edge conditions are tested

#### Flavor

Code is simple. Little state, very little control flow. 

Some code duplication for gas reasons.

Still unhappy about co-mingling principle and rewards funds.


#### Overflow

- [x] Solidity 8.0
- [x] Check that all for loops use uint256
  - [x] No loops!

#### Black magic

- [x] Does not contain `selfdestruct`
- [x] Does not use `delegatecall` outside of proxying
- [x] Address.isContract should be treated as if could return anything at any time, because that's reality.
- [x] No unbounded memory function arg onto storage

#### Authentication

- [x] Never use tx.origin
- [x] Check that every external/public function should actually be external
- [x] Check that every external/public function has the correct authentication
  - [x] No authentication. Staking pulls funds from the sender. Unstaking and Extend work off the sender.

#### Cryptographic code

_No crypto_

#### Gas problems

_No forloops_

#### External calls

- [x] No calls to external user sent contract addresses
- [x] Unsafe external calls
  - [x] Only calls internal contracts
- [x] Malicious behaviors
  - [x] Only calls internal contracts. (_Would be vulnerable to errors in reward source_)
- [x] Could fail from stack depth problems (low level calls must require success)
  - [x] No low level calls
- [x] No slippage attacks (we need to validate expected tokens received)
  - [x] No trading
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent
  - [ ] If oracle can be bent, it won't hurt us.
- [x] Safe balanceOf on for external contracts?
  - [x] No external balanceOf calls


#### Ethereum

DanielVF:
Quick illustration of the inflation slopes used by the RewardsSource contract

DanielVF:
#### Requirements

The RewardsSource contract mints tokens according to a schedule and sends it to another target.

The inflation rates are contract are configurable.

### Deployment Considerations

- [ ] We'll probably place this contract behind a proxy contract.
- [ ] Verify max inflation per day is the correct amount

#### Internal State

`ogv` Set in constructor. Only used to mint OGV from when sending rewards.

`rewardsTarget` Can only be set by Governor. Only used to verify collect collectRewards() caller and to send rewards to this address. 

`lastRewardTime` Set in constructor to current time. Set after rewards collection to current time. Used to calcuate the rewards duration and to skip rewards if we have already done so this block. 

`currentSlopeIndex` Cache of current position in the slopes list. Allows the common case to be a single slope lookup. Is reset to zero when slopes are replaced. Is written to when it moves forward. Read when starting to calculate inflation.

`inflationSlopes` contain the configurable inflation rates. Only set by the governor. Only read when calculating rewards. Slope start times must increase. Slope rates must be under the max inflation per day. End time is ignored.

- 🔴 Was missing a line for checking slope time increases. ✅ Corrected.

#### Attack

If lastRewardTime were massivly in the past, it could mint a large amount of funds money. This should be safe because is only set to the current timestamp. Miners do not have the ability to decrease timestamps below the previous block.

If rewards were given out and lastRewardTime was not set to now, we'd mint increasingly larger rewards. It's set right next to the mint.

If the end time on a slope segment was incorrectly set into the future, it could double mint over a time period. The logic here looks correct, though it could be better expressed in the code.

The rewards to slope math must be correct. The two most important parts are the initial range set, and trimming the ends of the range to the slope. This looks right.

#### Logic

- Might be good to run collectRewards when calling setInflation, so that all past rewards are paid at the scheduled rate. Would also protect against accidentally setting against a large past inflation amount.

#### Tests

  - [x] Each publicly callable method has a test
  - [x] Each business if branch has a test
  - [ ] Each logical branch has a test
  - [ ] Each require() has a test
  - [ ] Edge conditions are tested

#### Events

- :orangecircle: Missing events for setting config

#### Flavor

_The flavor is good._

#### Overflow

- [x] Check that all for loops use uint256

#### Black magic

- [x] Does not contain `selfdestruct`
- [x] Does not use `delegatecall` outside of proxying
- [x] Address.isContract should be treated as if could return anything at any time, because that's reality.


#### Authentication

- [x] Never use tx.origin
- [x] Check that every external/public function should actually be external
- [x] Check that every external/public function has the correct authentication

#### Cryptographic code

_No crypto_

#### Gas problems

- [x] Contracts with for loops must have either:
    - [ ] A way to remove items
    - [x] Can be upgraded to get unstuck
    - [x] Size can only controlled by admins
    - [x] Size limited to a reasonable max
- [x] Contracts with for loops must not allow end users to add unlimited items to a loop that is used by others or admins.

#### External calls

_The only external call is to ogv._

#### Ethereum

DanielVF:
Audit fixes checklist:

- [x] H-01 #88 
- [ ] H-02
- [x] M-01 #89
- [ ] M-02 Events
- [ ] M-03 Tests
- [ ] L-01 Docs
- [x] L-02 #93 
- [x] N-01 #86 #87
- [ ] N-02 Gov Numbers (Tom)
- [ ] N-03 Imports
- [ ] N-04 License identifier

franckc (approved):
LGTM - let's merge and then create separate PRs if needed to address any further comments from the audit.

PR #82, opened 2022-05-06
WIP: Claim page

tomhirst:
Getting a head start on this while we figure out the final product requirements, cc: @micahalcorn 

The initial layout provides a claim page with basic UI for the two scenarios detailed by @joshfraser in Discord:

> 1. OGN holders are eligible to claim unlocked OGV and have the choice to lock it for a duration of their choosing
>
>2. retroactive rewards to OUSD holders will be claimable in the form of pre-locked OGV that vests annually over 4 years. ie. they should be able to claim veOGV for the merkle distributor that entitles them to claim OGV after 1/2/3/4 years.

Screenshot 2022-05-06 at 11 55 41

To do:

- [ ] Integrate with contracts when product requirements finalised
- [ ] Build UI to satisfy the resulting claiming scenarios


tomhirst:
I've put together a v0.1 of this based on product discussion with @micahalcorn and the spec here: https://docs.google.com/document/d/1ediLk-ELVuvu-2PLZKPQnhyr4ihcem29J5-uMFmn1NU/edit#

It'd be good to get some eyes on this to make sure we're heading in the right direction. Most functionality is mocked at the moment, with my primary focus being on UI as we're laying this out sans-design input.

We discussed making the claim experience staged. My first instinct was to build out a simple wizard with a step tracker to take the user from education about OGV and the claim, through wallet eligibility and finally, the claim itself. Though, we can simply move each step to separate pages if we want to.

**Step 1 - Explanation**

We are awaiting content here.

Screenshot 2022-05-13 at 10 04 14

**Step 2 - Eligibility**

This step lets the user connect their wallet to see what tokens they're eligible to claim. The UI will also show the eligibility criteria and what tokens users can expect to receive for meeting each criterion listed.

Screenshot 2022-05-13 at 10 06 30

Screenshot 2022-05-13 at 10 06 52

Screenshot 2022-05-13 at 10 08 37

**Step 3 - Claim**

This step is where the user finally claims their tokens and we encourage them to lock up. The centre panel is where the user performs actions, the adjacent panels are informational. To the left, I think it'd be smart to include a list of benefits for locking up the to be claimed OGV right away. To the right, we have stats and recent transactions conveying social proof.

In the spec, we had discussed adding the users current balances and current lockups. I'm not sure this makes sense for the claim experience as I assume it's unlikely that users claiming their OGV/veOGV will already have some? So I think we can add a current lockups panel to the Vote Escrow page instead.

Screenshot 2022-05-13 at 10 16 44

All comments welcome!

joshfraser:
This is a great start. I'll let @micahalcorn weigh-in and have the final say, but here are some immediate reactions:

- I think we should show the eligibility on the first screen. People often use multiple wallets and will want to be able to quickly check them to see if any of them are eligibile before stepping though the whole flow.
- Once we have an eligible wallet, we will have a captive audience as we walk them an introduction to each of our tokens and how they work. this is the best opportunity we've had in a long time to educate people about each of Origin's products. let's not waste this opportunity.  I think we break these out into separate tabs:
  - OUSD (what it is & why it's awesome)
  - OGV (the governance token for OUSD)
  - veOGV (the longer you lock, the greater your rewards!)
  - OGN (OGN staking is coming/here! maybe we actually start with this so we don't break up the flow)
- the claim page looks a bit cluttered. we want people to understand how vote locking works & why they should lock up before they land on this page.
 - you can find the logos for OGV and veOGV here: 
   - https://github.com/OriginProtocol/origin-website/tree/master/static/img/coins
 - nit, let's left-align the ✅ and ❌ icons
 - all the veOGV is going to unlock annually over 4 years (4 separate locks). we'll want to communicate this visually.
   - For example, if we lock 400 OGV, we should show 4 bars with 100 OGV locked in each one. however, given the way voting power decays over time this equates to 250 veOGV. we probably want to show both numbers:
     - 25 veOGV on the 1 year lock
     - 50 veOGV for the 2 year lock
     - 75 veOGV for the 3 year lock
     - 100 veOGV on the 4 year lock
- we should include a big staking rewards APY indicator next to the slider that dynamically updates as you slide it (triple-digit APY if you lock for 4 years, or a low single-digit APY if you only lock for a week)
- we might not end up including the average lock time. it's less useful now that we have vesting vs. 1 single lock.
- low priority, but for the latest activity feed it would be fun to animate these in real-time (or replay the last 10 people who locked up) to really drive home the impression that EVERYONE is locking up.  
  - "joshfraser.eth just locked 5,000 OGV for 4 years [ethlink]"

Might have more later thoughts but wanted to share some quick feedback. Appreciate you taking the initiative on this!

micahalcorn:
Thanks for taking the first pass at this @tomhirst 👍 

@joshfraser I imagined that the eligibility and claim steps would be reactive to changes in the wallet. If so, checking a different address or claiming from one wouldn't require the user to go through the educational content again. I would think that we would want everyone to go through that whether they're eligible to claim or not.

I would leave the claim transactions out of the recent history and just include the ones that involve locks. We wouldn't want the user to see that the five most recent people claimed and didn't lock.

Visually, I would prefer to use a grey ✖️ rather than the red one (we can get one from @auregimon). Red looks sad and scary. I think of these more as disabled line items that are not relevant to you as opposed to negatives. I'm also not really sure what I'm looking at on the right side (number and grey circle). Does the circle change colors or is it a placeholder for the token icon? I'm a little conflicted on whether or not the eligibility form should reflect the post-claim status since that should probably be covered on the next screen and that indicator might otherwise be confused with the eligibility indicator. Along those same lines, I would probably change the heading to just "Tokens" instead of "Claimable tokens" since they might have been claimed if the user returns to this step.

I know it's difficult to visualize the educational parts at the beginning without having some content. We probably want those to be multiple concepts that you click through or scroll through, so it might make sense to have some vague progress dots like an Instagram gallery instead of three distinct steps where it looks like you could skip straight to the second or third.

I'll also circle back and add some other thoughts later.

tomhirst:
Thanks for your comments on this so far @micahalcorn @joshfraser. Here's a second run, with a focus on decluttering and simplification. I have a few queries too.

**Step 1 - Explanation**

Screenshot 2022-05-18 at 16 11 04

Screenshot 2022-05-18 at 16 11 13

- This step has been renamed Origin Products as it makes more sense from a user perspective re: the content to follow
- I've built out a tab-based system that lets us walk the user through each product individually, as suggested
- We disable progress to the Eligibility step until each product tab has been cycled through (and hopefully read!)

_Queries:_

1. Do we have a timeframe for the educational content? The diagrams on https://www.ousd.com are great, it might make sense to include something similar here

**Step 2 - Eligibility**

Screenshot 2022-05-18 at 16 11 37

Screenshot 2022-05-18 at 16 08 33

Screenshot 2022-05-18 at 16 12 01

- In agreement with the comments about displaying claimable amounts in both the eligibility and claim steps, I've removed the figures here. I don't think it's needed to show the number of tokens here because it's shown in the next step. Plus, more people will actually claim once they get to the final step
- I've softened the ineligible wallet state. The red cross emojis, while placeholders, were too negative. We could replace the emojis with softer (grey) icons, but I don't think we need them at all. Because, as @micahalcorn mentioned, these criteria are not something the user has missed out on, simply something they don't qualify for
- The eligible and ineligible status messages are clearer at a glance now
- Icons for the tokens have been added
- A user can only progress to the Claim step if they have a connected wallet that is eligible to claim OGV and/or veOGV

**Step 3 - Claim**

Screenshot 2022-05-18 at 18 45 11

Screenshot 2022-05-18 at 18 47 42

Screenshot 2022-05-18 at 18 48 21

- The benefits column has been removed. This helps declutter the step. I agree with @joshfraser that the benefits should ideally be apparent by this stage once the user has gone through the education step
- I've added veOGV and APY indicators to show the user what they're in line to get should they lock up their OGV. Though votes will decay, I think it's important to show the user what they'll get at the start (Curve do this, too). These will change based on the lockup length selected as the user slides. The button reads "Claim and Lock" with any length above 0 weeks. With 0 weeks selected, it reads "Claim". This pattern should encourage lockups
- The labels beneath the slider are now clickable, providing a shortcut for the user to common lockup durations
- I've mocked up a bar chart visual, showing two data sets against each other across the available lockup periods. The first currently depicts OGV staked and the second veOGV received based on the example figures provided
- I've cleaned up the OGV statistics column based on your feedback

_Queries:_

1. The staking contract's stake function always asks for an amount. Are we looking to allow the user control to set this with another slider between 0 and their claimable OGV here, or will we pass their max amount claimable as an assumed default? My take is it might be better to default to max claimable, but have this adjustable with a slider, too.
2. I like @joshfraser's idea about animating the latest events. We can absolutely cycle through these in their current state in the side column. But it might be better to display them as individual social proof pop-ups, ala https://fomo.com/product-tour. Does something like this fit in with our broader set of UI/UX guidelines?
3. @joshfraser I could use an explainer on the envisaged voting decay mechanism so I can properly feed the bar chart. Which contract(s) should I be calling to grab this information? cc: @DanielVF @tomlinton 
4. The claim of veOGV feels tacked onto the end of the OGV experience. Is there any good reason why we have to include this alongside the OGV claim? Could this be a separate experience, preventing confusion and decluttering the claim step?

It might be a good idea to further workshop this together soon @micahalcorn.

Cheers! :handshake:

tomlinton:
> My take is it might be better to default to max claimable, but have this adjustable with a slider, too.

Yea agreed this sounds like a good default.

> I could use an explainer on the envisaged voting decay mechanism so I can properly feed the bar chart.

This is not known ahead of time as it is a function of staking activity. My understanding might be a little foggy on exactly what we've settled on here, but here is my take:

- A bar chart might not be the best fit. Our veOGV implementation no longer decays, the voting power you hold just gets diluted. You'll always hold the same amount of veOGV, but your voting power is proportionally less with subsequent lockups. So the veOGV bar will always just be constant.
- Rewards are in the form of OGV. From my understanding of the bar chart it is saying rewards are paid out in veOGV.

I think in place of the bar chart we might say something like "You'll have 25 veOGV. Over the period of your lockup you'll earn xyz OGV (based on current APYs)."

> The claim of veOGV feels tacked onto the end of the OGV experience. Is there any good reason why we have to include this alongside the OGV claim? 

I don't think we have to. My first instinct is to simplify the whole thing and move the panel up alongside the OGV claim panel and have a single claim button. Then make "Lockup length" read something like "Lockup your 100 OGV". The text explainer from above might then become "You'll have 25 veOGV from your lockup and 50 veOGV. Over the period of your lockup you'll earn xyz OGV (based on current APYs)."

Maybe a tabular thing might work:

```
You are claiming:

veOGV eligibility:  50
veOGV from lockup: 25
OGV: 0
Projected OGV rewards:  200
```

And when the lockup length slider changes the figures change.

As you can see I'm not a good UI designer.

tomhirst:
I wanted to explore another idea before finalising the UI scope in our follow up call tomorrow, based on @joshfraser's document and the discussions we've had in the last few days.

TLDR: 

I think tables are the best way to explain to the user exactly what they are claiming and what they're locking up. Especially for those who're claiming OGV. Here's a walkthrough of what I've been looking at:

**Default state**

Screenshot 2022-05-30 at 18 47 14

- User is eligible for 100 OGV
- Lockup reward APY is brought to the forefront alongside the amount claimable
- Their lockup defaults to the max lockup duration
- The token being received (veOGV) is clearly shown
- We calculated what the projected OGV returned as rewards will be with this configuration
- The token being locked up (OGV) is clearly shown, with an unlock date based on the lockup duration selected in the slider

Discussion point: Should we allow users to set the amount of their eligible OGV that they want to lock up? Or, do we want to encourage an all or nothing scenario during the claim phase, with the ability to be flexible later, in the Vote Escrow experience?

**Reduced lockup duration state**

Screenshot 2022-05-30 at 18 47 25

- veOGV amount adjusts based on the vote power mechanism described in the scope document here: https://docs.google.com/document/d/1L5FhyEUg3n73p1PBW9J4LQxyAa1oKuBoIY-uED_1DwM/edit
- As does lockup reward APY and project OGV rewards (I'm using pseudo-math here at the moment until we start interacting with the contracts; take these figures as examples only)

**0 week lockup duration state**

Screenshot 2022-05-30 at 18 47 36

- OGV (not veOGV) is shown as the token to be claimed
- Lockup reward APY and projected OGV rewards remain and are shown as 0
- A "loss aversion" note is displayed to show users what rewards they are missing out on should they choose to not lock up

**Translating this to the veOGV claim panel**

Our original idea was to use a bar chart to show the difference between the OGV totals and veOGV totals over the 4 individual lockups (1 year, 2 years, 3 years and 4 years).

However. With the table idea, we can show this a little more explicitly, as demonstrated below:

Screenshot 2022-05-30 at 19 07 19

The negative is that the list is quite lengthy, but I think this approach helps to show the user exactly what's happening when they claim.

joshfraser:
Let's not show "Projected OGV rewards" as that's overly misleading. While we don't necessarily want to draw attention to the fact that rewards will be dropping over time, we don't want to lie about what users should expect either.

tomhirst:
> Let's not show "Projected OGV rewards" as that's overly misleading. While we don't necessarily want to draw attention to the fact that rewards will be dropping over time, we don't want to lie about what users should expect either.

Understood. I thought this was something we could work out pretty definitively from the contracts/current yield on offer. Easily removed in the next round.

tomhirst:
Claim layout revised to a single column with global stats on top:

Screenshot 2022-06-01 at 17 23 59

tomhirst:
Automated global lockup alerts UI (will have ENS lookup on mainnet):

Screenshot 2022-06-02 at 11 02 16

micahalcorn:
This is great @tomhirst, thanks. If I understand correctly, this is a single quiz for all of the content? It seems like most users would be inclined to just scroll to the bottom and start answering the questions. Ideally, we would break each concept up into a different step with a dedicated quiz and then put a speed bump on each one (like a timeout or something on the continue button). As for the layout, I don't have a strong preference for whether it is single-column vs the sticky panel. I like the idea of the user having to scroll past the content to get to the quiz instead of it being right there at the top, but I also like that the user is already at the top of the content when she's looking at the quiz and realizing that she needs to read.

We might want to add some static graphics that represent each concept or it might make sense for those to be the posters on the videos.

tomhirst:
> This is great @tomhirst, thanks. If I understand correctly, this is a single quiz for all of the content? It seems like most users would be inclined to just scroll to the bottom and start answering the questions. Ideally, we would break each concept up into a different step with a dedicated quiz and then put a speed bump on each one (like a timeout or something on the continue button). As for the layout, I don't have a strong preference for whether it is single-column vs the sticky panel. I like the idea of the user having to scroll past the content to get to the quiz instead of it being right there at the top, but I also like that the user is already at the top of the content when she's looking at the quiz and realizing that she needs to read.
> 
> We might want to add some static graphics that represent each concept or it might make sense for those to be the posters on the videos.

Yep, that's how it is right now, but we can change it. Question:

Do we add each educational stage (OUSD, OGV, OGN) as their own "page" stages (like Check eligibility,  Learn about Origin and Claim airdrop) are now? Or, do we have them as inner stages within the "Learn about Origin" page?

We wouldn't need to add a time limit to activating any progress buttons as we'd only activate progress to the next stage (in either of the above scenarios) by completing the current quiz correctly.

Re: single-column vs. sticky panel. Visually, I prefer the sticky panel, but my product head says we keep it at the bottom to enforce the content being read pre-quiz interaction as much as we can. So let's go with that.

If we get the videos done, posters make sense as an additional graphic might be overkill. If we don't get the videos done, we'll definitely need some graphical elements to break up the text instead.

tomhirst:
Screenshot 2022-06-15 at 11 57 05
Screenshot 2022-06-15 at 11 57 13
Screenshot 2022-06-15 at 11 57 34
Screenshot 2022-06-15 at 11 57 47
Screenshot 2022-06-15 at 11 58 00
Screenshot 2022-06-15 at 11 58 07
Screenshot 2022-06-15 at 11 58 22

@micahalcorn Here's the latest quiz UX. We now have separate sub-steps for each product we want to educate on, separate quizzes with different questions for each and a sub-section sidebar that tracks the sub-step progress.

In the last sub-step, after the quiz is passed through correctly, the final claim step becomes accessible.

tomhirst:
Screenshot 2022-06-15 at 12 08 08

Screenshot 2022-06-15 at 12 08 29

@micahalcorn Here're some mobile view screenshots.

tomhirst:
> It also might be good to change the treatment on the back button so that it's not as equally weighted as the thing we expect you to want to do. I'm a little confused when I see several big green buttons near one another.

Thanks, both of these points make a lot of sense. Have actioned, screenshots below:

Screenshot 2022-06-18 at 15 19 37

I added a slight top offset to the small sidebar to emphasise its presence as supplementary to the main content and retain nice spacing on scroll.

Screenshot 2022-06-18 at 15 19 46

tomhirst:
@sparrowDom Thanks for merging. Looks like everything is there, just had to fix the layout logic based on the claim open timestamp.

I've added some progress to the multi-lock support here too, which now works to a basic level.

I think the aim should be to get this branch into master/stable as soon as possible. Probably once we have the merkle contracts tied into the claim UI.

tomhirst:
Closing as got too large and merged as part of #113

PR #88, opened 2022-06-01
OZ H-01 Add missing collectRewards() to extend()

DanielVF:
Without this fix earned rewards would dropped during extension.

This also includes a change in the way the rewards debit is accounted for that allows it to be handled entirely inside the collectRewards functions, rather than needing it's own line of code at the end of each function that does something with a user's account.

### DIFF
--- a/contracts/OgvStaking.sol
+++ b/contracts/OgvStaking.sol
@@ -21,7 +21,7 @@ contract OgvStaking is ERC20Votes {
     // 3. Reward Storage
     ERC20 public immutable ogv;
     RewardsSource public rewardsSource;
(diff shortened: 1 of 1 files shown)


DanielVF:
This PR contains a vulnerability, in that if the totalSupply is zero the user's rewardDebtPerShare is not set.

This would cause a problem if all users were to unstake, the totalSupply were to drop to zero, and then a user, who was not the last user to unstake, waits some amount of time after their unstake, and restakes. This results in their rewardDebtPerShare still being set to the old value, meaning that when they next collect rewards, they would get too many rewards.

image

DanielVF:
Fixed in the above PR.

PR #90, opened 2022-06-02
Retroactive airdrop calculations

tomlinton:
This is to a point where we can begin to validate the data.


retroactive/src/ousd.ts
```diff
@@ -0,0 +1,160 @@
...
+      handleOusdTransfer(blockNumber, event);
+    } else if (event.to === wousdContract.address) {
+      handleWousdTransfer(blockNumber, event);
+    }
+  }
+
+  if (blockNumber) {
+    process.stdout.write(
+      `${blockNumber} - ${Object.keys(ousdHolders).length} OUSD holders, ${Object.keys(wousdHolders).length
+      } wOUSD holders\r`
```
sparrowDom, line 85: This carriage return progress log is pretty cool

sparrowDom (commented):
Left a couple of comments PTAL

retroactive/src/ousd.ts
```diff
@@ -0,0 +1,160 @@
...
+    fs.writeFileSync("ousd-claims.json", JSON.stringify(claims));
+  }
+
+  done();
+});
+
+const calculateRewards = () => {
+  console.log("\n");
+  console.log("Calculating OUSD rewards");
+  const ousdRewards = rewardScore(ousdHolders, SNAPSHOT_BLOCK);
```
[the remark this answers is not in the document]
tomlinton: Fixed the block to the redeploy 👍

retroactive/src/ousd.ts
```diff
@@ -0,0 +1,160 @@
...
+  console.log("Calculating wOUSD rewards");
+  const wousdRewards = rewardScore(wousdHolders, SNAPSHOT_BLOCK);
```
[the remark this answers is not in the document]
tomlinton: The event listener starts from the OUSD redeploy block so it should naturally include all events since WOUSD launch.

retroactive/src/utils.ts
```diff
@@ -0,0 +1,138 @@
...
+    return value;
+  } else {
+    return BigNumber.from(value);
+  }
+};
+
+const rewardScore = (
+  addressHistory: { [address: string]: BlockHistory[] },
+  stopBlock: number,
+  rewardFromBlock = 0
```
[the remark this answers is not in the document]
tomlinton: Yea this is right, I forgot to handle this as its own unique case

retroactive/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "ogv-retroactive-airdrop",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
+  "dependencies": {
+    "@types/lodash": "^4.14.182",
+    "eslint-config-prettier": "^8.5.0",
+    "ethereum-events": "github:AleG94/ethereum-events#8c92ade1d8f489a577399d5d3cd0093ef6c58aa5",
```
DanielVF, line 9: Note that I had to change this to https://github.com/AleG94/ethereum-events.git#8c92ade1d8f489a577399d5d3cd0093ef6c58aa5 to get it to install for me.

retroactive/src/ousd.ts
```diff
@@ -0,0 +1,183 @@
...
+} from "./utils";
+import {
+  ousdContract,
+  wousdContract,
+  OUSD_DEPLOY_BLOCK,
+  WOUSD_DEPLOY_BLOCK,
+} from "./contracts";
+import { ethereumEventsOptions, web3 } from "./config";
+
+const AIRDROP_AMOUNT = 400000000;
```
DanielVF, line 18: Not sure if this is intentional or not, but this airdrop amount here is not in 1e18 fixed point.

This means that:
- the math calculations clip, rounding down to the nearest whole number.
- the CSV output in in non-fixed point, whole numbers only.
- 2/3s of the CSV output is "0,0,0" on the amounts.

DanielVF:
@tomlinton Do we need to explicitly exclude Curve and friends?

I believe the plan with the airdrop was to exclude those who providing earlier liquidity, since most of those were eligible for previous drops. Currently, Curve, uniswap, et all are included the in the airdrop calculation script (though obviously unable to claim), and rather dominate it. In this script curve makes up 86% of the entire airdrop. Add in uniswap v2 and v3, and we are up to 94%. Add some other big contracts (flipper, compensation, etc) that will be unable to claim (plus Kucoin), and everyone else in the system is dividing up only 1.42% of the pie.

image

(I'm also guessing that there's a bug somewhere in the calc code if contract that swaps a lot, it gets boosted. The above numbers are off from what I would expect, although curve would dominate regardless.)

retroactive/src/utils.ts
```diff
@@ -0,0 +1,160 @@
...
+  if (BigNumber.isBigNumber(value)) {
+    return value;
+  } else {
+    return BigNumber.from(value);
+  }
+};
+
+// Calculate a reward score based on the holdings of a token multipled by the
+// number of blocks it was held for.
+const cumulativeRewardScore = (
```
DanielVF, line 29: I'm reasonably sure there's a bug in this method. Looking at the score data, swap related contract are relatively "outperforming" holders.


I don't know if this is the same bug or not, but contract 0xa79828df1850e8a3a3064576f380d90aecdd3359 is a deposit contract that never actually holds OUSD outside of an in-out transaction. And yet it's marked as the 7th biggest receiver of an airdrop.

image

DanielVF: Somehow it's calculating every single balance forward as if it's held all the way to the snapshot block.

image

image

DanielVF: Bingo: 
image

DanielVF: Also need to drop skipping the first entry and change the non-end block to something more like

```javascript
lastBlock = blockNumber;
firstBlock = Math.min(stopBlock, history[currentIndex + 1].blockNumber)
```

tomlinton: I think it works if we remove skipping the first history entry and leave the rest as is.

`Math.min(stopBlock, history[currentIndex + 1].blockNumber)` should always just resolve to stopBlock because the history stops at stopBlock (i.e. we don't query events passed that point.

Am I thinking about that correctly? 😅

DanielVF: Doesn't feel like that's correct.

DanielVF:
The current code is wrong - you can sanity check by seeing all the negative numbers in the output.

retroactive/src/ousd.ts
DanielVF: Pinging on this one.

retroactive/src/utils.ts
DanielVF: Looks roughly good now.

DanielVF (commented):
I pulled several accounts, and used a spreadsheet to manual calculate balances in and out, and compare to the ousd.ts output. Everything looked accurate to three decimals.

retroactive/src/ousd.ts
tomlinton: Updated this 👍

sparrowDom:
@tomlinton FYI you have in TODO to generate merkle root. With `claims.json` provided this script by @tomhirst already generates the Merkle root & proof for every account: 
https://github.com/OriginProtocol/ousd-governance/blob/master/package.json#L8

sparrowDom:
@tomlinton about how we handle OUSD. Are we intentionally ignoring the rebasing logic?

tomlinton:
> @tomlinton about how we handle OUSD. Are we intentionally ignoring the rebasing logic?

Yea, the events emitted don't really give us enough information to handle it nicely.

tomlinton:
> I think it should be possible enriching transfer events with rebase events. It does increase complexity of the code of course... I have done something similar in this Dune query: https://dune.com/queries/975795?Address_t6c1ea=0x9efab36288565dC343b6909350767cb392AA4F69&score_adjustment_t6c1ea=255159.22
> 
> 
> 
> I would expect if we were to compare rebasing scores with non rebasing scores for results to differ up to 15-20% ballpark somewhere. 

I'm happy to just leave it out. Without rebasing we are rewarding OUSD buys which is a fairer than including the yield which was often not that smooth.

sparrowDom:
I would argue that it is good to be the closest to the "truth" as possible and that would be including the rebases. But after inspecting the actual difference between rebased / non rebased rewards I was surprised by how small they are. Because of that I don't think it is worth the hassle of adding additional complexity to the code. 

Also **good news** I've made a [Dune query](https://dune.com/queries/975795?score_adjustment_t6c1ea=277900.22&ogn_rewarded_t6c1ea=400000000000000000000000000&totalScore_t6c1ea=111711574716639470472974186736010&Address_t6c1ea=0x37Facc790b36Dc08446381c4873962F2BC94A5D2) that makes an ogv calculation including and excluding rebases and the data almost completely matches what your script produces @tomlinton . Results [are here](https://docs.google.com/spreadsheets/d/1HKbqdx0VK_4SiHriAkEY9TMcdRNTyTBXONOluz3tfRg/edit?usp=sharing).
In the 20 handpicked accounts there is up to 3% difference in my approach, but I think those differences are negligible.

tomlinton:
@tomhirst @sparrowDom This branch contains a Rinkeby deploy and all the airdropdata up to a fairly recent block. I think we should merge it to master. The OGN/LM script may not be correct yet and needs more review, but we can fix forward.

tomlinton:
Merging and fixing forward

sparrowDom:
sweet thanks a lot @tomlinton

PR #99, opened 2022-06-06
Holding page

tomhirst:
Screenshot 2022-06-06 at 14 53 10

Here's the first run at a holding page for OGV, including:

- Countdown timer to airdrop date
- Key dates so users can see what to expect from the process
- CTA buttons for obtaining OGN and OUSD
- A link to the [medium article](https://blog.originprotocol.com/tokenomics-retroactive-rewards-and-prelaunch-liquidity-mining-campaign-for-ogv-1b20b8ab41c8) about the OGV token launch
- App reverting to live structure once `CLAIM_OPENS` timestamp has passed

To do before this can be merged:

- [ ] Content review cc: @micahalcorn @joshfraser 

Then...

- [ ] Decision on where this will be deployed (do we have a domain?)
- [ ] CI/CD setup cc: @sparrowDom @franckc


sparrowDom:
let me take care of CD/CI

micahalcorn:
> * [ ]  Content review cc: @micahalcorn @joshfraser

Let's change the second button to "Be an OUSD LP" and link it to the Curve pool. I like the focus on the two actions that we want them to take rather than my proposed focus on learning more.

Also, I'm wondering if the dates will be specific to the user's time zone. It would probably be odd if they don't literally match the blog post, but dates and times are usually client-specific in a case like this.

tomhirst:
Screenshot 2022-06-14 at 11 28 03

@micahalcorn Latest visual changes.

micahalcorn:
> @micahalcorn Latest visual changes.

I like it, thanks!

tomhirst:
Screenshot 2022-06-14 at 19 44 11

@micahalcorn Mobile view.

sparrowDom (approved):
LGTM

PR #101, opened 2022-06-07
Merkle distributor contracts (optional/mandatory)

tomlinton:
Draft of the optional and mandatory lockup distributor contracts. @sparrowDom @franckc said you might be interested in working on these. We need to add a shutdown state to these contracts where we can reclaim any OGV on the distributor contracts after the claim period has expired.

Edit (Domen):
- added shutdown state (burning remaining OGV) once the claim period expires. 

**Related issue:** https://github.com/OriginProtocol/ousd-governance/issues/109


tomhirst:
FYI @tomlinton added a fix for the deploy script, so I can merge this into #82 and start integrating the contracts with the UI. Hope that's OK!

tomlinton:
Yep no problem, I think I broke it with a bad merge

contracts/MandatoryLockupDistributor.sol
```diff
@@ -0,0 +1,94 @@
...
+        token = _token;
+        merkleRoot = _merkleRoot;
+        stakingContract = _stakingContract;
+    }
+
+    /**
+     * @dev
+     * @param _index Index in the tree
+     */
+    function isClaimed(uint256 _index) public view returns (bool) {
```
DanielVF, line 41: @tomlinton What's the purpose of this index bitmasking dance vs just reading and writing a boolean to a mapping?

DanielVF:
Looks like these are missing the mentioned OGV recovery methods after shutdown?

contracts/MandatoryLockupDistributor.sol
tomlinton: It's a fork of the uniswap merkle distributor which seems easier than rewriting it. I'd guess it's also more gas efficient.

tomlinton:
That can be separate PR, for now I just want to merge this so we can get a Rinkeby setup going.

sparrowDom:
@tomlinton feel free to merge this. Have added a burnable function but have not tested it. Will do any fixes in subsequent PRs.

contracts/GovernanceToken.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.4;
 
-import "OpenZeppelin/openzeppelin-contracts-upgradeable@4.6.0/contracts/token/ERC20/ERC20Upgradeable.sol";
```
DanielVF, line 4: I see the change to the import, but does this actually get used?

sparrowDom: My bad forgot to add it to contract definition. Should be ok now

contracts/MandatoryLockupDistributor.sol
```diff
@@ -0,0 +1,58 @@
...
+        address _token,
+        bytes32 _merkleRoot,
+        address _stakingContract,
+        uint256 _endBlock
+    ) AbstractLockupDistributor(_token, _merkleRoot, _stakingContract, _endBlock) {}
+
+    /**
+     * @dev Execute a claim using a merkle proof with optional lockup in the staking contract.
+     * @param _index Index in the tree
+     * @param _amount Amount eligiblle to claim
```
DanielVF, line 28: Spelling: "eligible"

DanielVF:
Are tests passing for you guys? 

I'm getting a lot of failures: 

image

Looks like most are around making a proposal.

image

tomlinton:
Needs --network hardhat because it uses hardhat_mine calls

contracts/MandatoryLockupDistributor.sol
```diff
@@ -0,0 +1,58 @@
...
+     * @param _merkleProof The proof
+     */
+    function claim(
+        uint256 _index,
+        uint256 _amount,
+        bytes32[] calldata _merkleProof
+    ) external {
+        require(!isClaimed(_index), "MerkleDistributor: Drop already claimed.");
```
DanielVF, line 36: Just checking a design decision here. These claim methods don't stop working after the endBlock, it's currently more a soft end until we burn the funds and the staking calls break on balance transfer errors.


Do we want to just have a `require(block.number < endBlock)` in these claim methods, so the end requires no human action, and is explicit and precise?

sparrowDom: Sounds like a good proposal to me. Adding that in!

sparrowDom:
@DanielVF change done

tests/helpers.py
```diff
@@ -1,16 +1,24 @@
...
 DAY = 86400
 WEEK = 7 * DAY
+MAXTIME = 4 * 365 * DAY
+TOL = 120 / WEEK
+
 
 def approx(a, b, precision=1e-10):
     if a == b == 0:
         return True
+    print(2 * abs(a - b) / (a + b))
```
DanielVF, line 13: Did we mean to have these two prints in here?

tests/helpers.py
```diff
@@ -1,16 +1,24 @@
...
+    print(precision)
     return 2 * abs(a - b) / (a + b) <= precision
```
DanielVF, line 15: Notice that this method of computing differences will crash if both the actual and the expected are both zero. ( `a+b` will equal zero, which will cause a divide by zero error.)

This math does seem to work for small differences, > 3 decimals, but it gets weird for low decimal precision. The difference between 50 and 100 comes out to 0.66666, which probably isn't what you would expect from such a large difference.

tests/helpers.py
sparrowDom: not sure why those are there... tnx removing

tests/helpers.py
sparrowDom: good feedback thanks. have changed the function to: 
- always return false if expected / received numbers have different sign
- instead of a / b have renamed properties as `expected`and `actual`. This way the denominate using the `exptected` property and not the average between expected and actual. so 50 (expected) and 100(actual) will yield `2` as in 200%. Which is more correct IMHO

sparrowDom:
@DanielVF can we merge this in?

DanielVF (approved):
These contracts look good. Couple very minor things noted, but after those are fixed, we are good to merge.

tomlinton:
Added those final things thanks @DanielVF, will merge when tests pass.

DanielVF:
Rock On!

PR #117, opened 2022-07-01
Increase vote power required for proposal

### LINKED ISSUE #116: Governance min proposal amount seems too small?
Currently the threshold for submitting a governance proposal is 2,500 voting power. At launch time that would be 2,500 OGV locked up for a week, or 238 OGV locked up for four years. This seems to make the cost of being able to spam proposals be minimal. Do we need to increase this value?

tomlinton:
Closes #116


franckc (approved):
LGTM

PR #118, opened 2022-07-01
Decrease vote time to 5 days

### LINKED ISSUE #115: Governance voting time may be too long?
We currently have a seven day voting time, plus a two day timelock afterwards. That means that any governor based response is going to take nine days. This feels too long?

tomlinton:
Closes #115


contracts/Governance.sol
```diff
@@ -19,7 +19,7 @@ contract Governance is
         Governor("OUSD Governance")
         GovernorSettings(
             1, /* 1 block */
-            45818, /* 1 week */
+            32727, /* 5 days */
```
franckc, line 22: Assuming each block takes ~15sec to mine, isn't 5 days more like ~28,800 blocks?
```
5 * 24 * 60 * 60 / 15 =  28800
```

tomlinton: I used the value they'd used, which was ~13.2 seconds. Based off this https://ycharts.com/indicators/ethereum_average_block_time that looks about right, although it spiked recently.

franckc: Makes sense. My bad, I did not realize block time these days was lower than 15sec.

franckc (approved):
LGTM

DanielVF (approved):

PR #119, opened 2022-07-01
Add mainnet and testnet data + update endpoint

tomhirst:


tomhirst:
Closing because we did this work in #120 and that has been merged.

PR #121, opened 2022-07-04
Upgradable OGV staking contracts

DanielVF:
Upgradable contracts for OGV staking.

Notes:

Uses proxy code from OUSD. I much prefer this style - with upgrade logic in the proxy - over UUPS proxies.

No contract specific initializers needed, just initialize the the proxies to point at the base contracts, and we are good to go.

I've changed the forge tests and the brownie tests over to use the proxies as part of the testing.

### DIFF
--- a/contracts/OgvStaking.sol
+++ b/contracts/OgvStaking.sol
@@ -10,9 +10,9 @@ import {RewardsSource} from "./RewardsSource.sol";
 /// @author Daniel Von Fange
 /// @notice Provides staking, vote power history, vote delegation, and rewards
 /// distribution.
-/// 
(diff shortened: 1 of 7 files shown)


franckc (approved):
Verified that InitializeGovernedUpgradeabilityProxy is the same implementation as in OUSD.

Changes to OgvStaking.sol and RewardsSource.sol LGTM

tomlinton (approved):
Looks good, although I'd prefer we kept the same proxy code throughout the contracts. Consistency of the upgrade interface seems more important. We can roll with it though, as we are time pressured.

PR #130, opened 2022-07-05
Configurable minimum staking duration

DanielVF:
Allows minimum staking time to be changed at contract deploy time.

Keeps  minimum staking time to seven days in tests, so that we don't have to redo all our math and tests.

### DIFF
--- a/contracts/OgvStaking.sol
+++ b/contracts/OgvStaking.sol
@@ -16,6 +16,7 @@ import {RewardsSource} from "./RewardsSource.sol";
 contract OgvStaking is ERC20Votes {
     // 1. Core Storage
     uint256 public immutable epoch;
+    uint256 public immutable minStakeDuration;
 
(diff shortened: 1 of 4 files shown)


contracts/OgvStaking.sol
```diff
@@ -16,6 +16,7 @@ import {RewardsSource} from "./RewardsSource.sol";
 contract OgvStaking is ERC20Votes {
     // 1. Core Storage
     uint256 public immutable epoch;
+    uint256 public immutable minStakeDuration;
```
franckc, line 19: nit: document the unit is seconds

franckc (approved):
LGTM

contracts/OgvStaking.sol
DanielVF: Done!

tomlinton (approved):

Issue #136, opened 2022-07-06
Extend processing failure by the listener causes lockup deletion


franckc:
I had some OGV locked for 1 year on Rinkeby under my wallet 0xD85A569F3C26f81070544451131c742283360400
I clicked on the "Extend" button to extend my lockup.
The transaction was processed by the contract: https://rinkeby.etherscan.io/tx/0xc4788c0354522d685a1eedbbd10d2eaf7a6b9ddeea53621d67b7d985a949c26c

But my lockup disappeared from the UI after a refresh.

I did some digging and it looks like the DB row for my lockup was deleted from the DB. Here is my theory
1. Note that the contract re-uses the same lockup id when extending a lockup. See the [contract code](https://github.com/OriginProtocol/ousd-governance/blob/6c2f7c709fd7dddc02bb2812734e737088a1e1e9/contracts/OgvStaking.sol#L201).
2. Two events, are emitted using the same lockupId. First the Unstake, then the Stake
3. For some reason that I don't understand, the listener processed the Stake *before* the Unstake. Here are the logs
```
2022-07-06T04:47:58.706552+00:00 app[worker.1]: info: Got confirmed block
2022-07-06T04:47:58.713592+00:00 app[worker.1]: info:
2022-07-06T04:47:58.713594+00:00 app[worker.1]: Invalid `prisma.lockup.create()` invocation:
2022-07-06T04:47:58.713594+00:00 app[worker.1]: 
2022-07-06T04:47:58.713595+00:00 app[worker.1]: 
2022-07-06T04:47:58.713597+00:00 app[worker.1]: Unique constraint failed on the fields: (`lockupId`) {"clientVersion":"3.10.0","code":"P2002","meta":{"target":["lockupId"]}}
2022-07-06T04:47:58.717336+00:00 app[worker.1]: info: Removed lockup
```
4. The processing of the Stake event failed due to the unique constraint violation. And the Unstake caused the deletion of the stake
5. As a result my lockup was wiped out from the DB and the UI does not show it while it is still present on the contract.

franckc:
I think it might be that we don't have await in the listener when calling handlers so that could cause the events to get processed out of order
https://github.com/OriginProtocol/ousd-governance/blob/master/client/listener.ts#L132

@tomhirst @sparrowDom @tomlinton what do you think?

sparrowDom:
[...]

PR #137, opened 2022-07-06
Add awaits to the listener to prevent out of order events processing

franckc:
Fix for #136


tomhirst:
Looks like #136 (just merged) missed out `async` on the switch function call. Is this needed too @sparrowDom? If so, we can merge this also.

tomhirst:
Just tested and it seems we are good after merging #136, so I'm gonna close this.

PR #144, opened 2022-07-06
start refactoring contracts setup

sparrowDom:
Creating proposals is broken right now because of my change some time ago how contracts are setup. Previously contracts object was deconstructed and `abi` property was added to the contract. Which is very convenient to tie together abi + contract functionality. But it is not a perfect copy since it does not copy over things like `connect` function that allows to connect a signer. 

For that reason in this PR the contract object now separately contains contract and abi implementation.


sparrowDom:
closing in favour of: https://github.com/OriginProtocol/ousd-governance/pull/288

PR #146, opened 2022-07-06
add awaits so the right order of event processing is ensured

sparrowDom:
add awaits so the right order of event processing is ensured

**issue:** https://github.com/OriginProtocol/ousd-governance/issues/136


franckc:
I think we should probably add an await in the main loop calling the handler to avoid possibly processing blocks out of order. will send a PR...

PR #152, opened 2022-07-06
Listener: user WEB3_PROVIDER vs hardcoded value

franckc:
I double checked and there is an env var WEB3_PROVIDER defined in heroku both for the Rinkeby and Mainnet deploys.


tomlinton (approved):

sparrowDom:
awesome thanks!

PR #158, opened 2022-07-07
Fix missing imports

franckc:
I broke the build with #152 since I did not realize INFURA_ID, mainnetNetworkUrl and rinkebyNetworkUrl were used elsewhere.

This PR re-introduces mainnetNetworkUrl and rinkebyNetworkUrl but still gets rid of INFURA_ID since really I'd prefer to not have these keys committed in the code.

The docs for WalletConnect are here: https://docs.walletconnect.com/quick-start/dapps/web3-provider#required and it looks like it support either InfuraId or an rpc mapping.


sparrowDom:
LGTM thanks @franckc . Yes agreed we will need to do some cleanup

PR #159, opened 2022-07-07
Prototype

tomlinton:


contracts/OgvStaking.sol
```diff
@@ -13,10 +17,10 @@ import {RewardsSource} from "./RewardsSource.sol";
 ///
 /// The balance received for staking (and thus the voting power and rewards
 /// distribution) goes up exponentially by the end of the staked period.
-contract OgvStaking is ERC20Votes {
+contract OgvStaking is Initializable, OwnableUpgradeable, UUPSUpgradeable, ERC20VotesUpgradeable {
     // 1. Core Storage
```
franckc, line 21: We should allocate a slot gap here, right?

franckc (commented):
Thanks a ton for working on this upgrade! Took a quick pass.
I haven't review your note about the storage slots at https://gist.github.com/tomlinton/03c0f916bedece450c9cc0a9a7325048

contracts/OgvStaking.sol
tomlinton: ERC20VotesUpgradable has a big gap in ContextUpgradeable but there are a few initializable flags stored before that so I think 1 slot will collide with this before the gap is hit https://github.com/OriginProtocol/ousd-governance/blob/c894a651edf21aff2944be3eb2aaa94d374fa167/contracts/VoteLockerCurve.sol#L47 but it depends how things are packed.

From looking at the old VoteLockerCurve implementation we need a gap of 10 storage slots.

```
string 6 1
string 40 2
uint8 8 3
uint256 constant 0
uint256 constant 0
mapping 32 4
mapping 32 5
[] 32 6
uint256 32 7
mapping 32 8
mapping 32 9
address 10
```

We could jam a blank contract with a gap in before ERC20VotesUpgradeable to skip all the written storage slots from VoteLockerCurve since the first three inherited contracts are now identical.

I might be miscalculating here. I'd love to get a visual view of the storage but don't know of a tool/haven't had time to generate one.

tomlinton:
I'm not worried about the values yet - just want to make sure the storage will work out before cleaning it up. I moved the stuff from the gist into the comment above.

contracts/OgvStaking.sol
DanielVF: We can see the slots written during the first initialize here: 

https://dashboard.tenderly.co/tx/mainnet/0xc1493d425538b5fed552477235a8179d299af178905cf7e34111e5ae3790fa26/state-diff


|hex|slot|Notes|
|--|--|--|
|0x33|51|
|0xc9|201| New is mapping, should be no conflict|
|0xca|202| New is mapping, should be no conflict|
|0xcb|203|New totalySupply, set to decimals **CONFLICT**|
|0xce|206|In gap|
|0xd2|210|In gap|

DanielVF:
Changes I've made:

Moved those variables back to immutable. They don't need to change in all likelihood for the life of the contract. We can redeploy if they ever do.

Added a gap to skip already used slots.

Added code to block initialize in constructor. I think anyone could have initialized our implementation contract, then upgraded and self destructed it, thus permanently bricking all staking/unstaking/and voting. This kind of deadly gotcha is yet another reason I hate UUPS proxies.

Added code to set the owner() in initialize(). While the main proxy we are using already has an owner and would be upgradable, any other deploys would have been unupgradable because there would have been no owner.

franckc:
Thanks for all your work fleshing out this solution @tomlinton and @DanielVF !
Per discussion on discord, the plan is to use the new deployment vs taking the risk to do this upgrade.
Let's keep this PR open until after the launch though - just in case...

tomlinton:
Closing, can reopen if needed.

PR #161, opened 2022-07-07
unmock staking toasts

sparrowDom:
unmock staking toasts. 
We can decide whether to:
- show new stakes immediately
- or add them to array of recent stakes and show predictibly


client/utils/useHistoricalLockupToasts.js
```diff
@@ -0,0 +1,141 @@
...
+import { useStore } from "utils/store";
+import { toast } from "react-toastify";
+import { utils } from "ethers";
+import { truncateEthAddress } from "utils";
+
+const useHistoricalLockupToasts = () => {
+  const { web3Provider, contracts } = useStore();
+  // 13.13 seconds is average block time (on 7.7.2022) ~
+  const blocksToLookBack = 1300; // roughly 2 days
+  const SECONDS_IN_A_MONTH = 2629743;
```
tomhirst, line 12: We need to check that we use the same value across the project. See Discord comment: https://discord.com/channels/404673842007506945/758123420230615080/994609704477134878

client/utils/useHistoricalLockupToasts.js
```diff
@@ -0,0 +1,141 @@
...
+
+  /* because of the way event listener handler code does not get access to updated
+   * react component state. Separate useEffect is needed to handle populating the
+   * recent lockups array.
+   */
+  useEffect(() => {
+    if (!newEvent) {
+      return;
+    }
+    setRecentLocups([
```
tomhirst, line 53: Typo

client/utils/useHistoricalLockupToasts.js
```diff
@@ -0,0 +1,141 @@
...
+        events.map((event) => {
+          return {
+            shown: false,
+            rawEvent: event,
+          };
+        })
+      );
+
+      // subscribe for events only after past events have already been fetched
+      OgvStaking.on(stakesFilter, newStakeListener);
```
franckc, line 96: I forgot how subscriptions are implemented. Is it using websocket push or a periodic pull?

What does that mean from a connection to Alchemy standpoint? Will every client hold a connection?

Just pointing this out since we wouldn't want to run into Alchmey capacity issue. Especially during the initial surge of users trying to claim rigth after the launch.

client/utils/useHistoricalLockupToasts.js
sparrowDom: good point changing to: 2592000

client/utils/useHistoricalLockupToasts.js
sparrowDom: tnx fixed

sparrowDom:
@tomhirst the way the code works ATM is that it always shows the lockups from the last 2 days in the same drip logic as you had before (random 5-20s intervals). The thing to decide is when a new lockup arrives in real-time whether:
- to show it immediately
- to add it to the list that drips it in 5-20s intervals. 

Whatever option we go with, it wouldn't change that the existing lockups from 2 days would still be dripping in.

client/utils/useHistoricalLockupToasts.js
sparrowDom: very good point!

Since right now we use provider on HTTP protocol instead of WS I would imagine these are periodic pulls. A thing to consider is that currently these happen only on the final step of the Claim page. Meaning it is a page where users won't spend super large amounts of time.

We could also disable the listener and not get real-time staging events if the concern still persists. You think it could be that impactful?

franckc: If it happens only on the last Claim page I think it should be ok.
Is there a way to make sure we cancel the listener when the user moves to a different page? So that we don't keep a bunch of listener actives if the user is not on that page.

tomhirst:
@sparrowDom In that case I think it's good to show them right away. So long as we don't have a situation where there are way too many toasts or way too few.

sparrowDom:
Cool, than leaving it as is.

Fixed the bug, with the too many shows of claim toasts. Mainly the fix was moving claims to global state, so the dapp persists which have been already shown.

tomhirst:
LGTM

Issue #184, opened 2022-07-08
Redirect to claim experience


micahalcorn:
The production environment currently does not redirect from / to /claim. The only route that should be accessible outside of /claim is /vote-escrow. This will change as we roll out the ability to create proposals.

Issue #171, opened 2022-07-08
Support countdown override


micahalcorn:
Let's use a localstorage variable or something similar to allow ourselves to circumnavigate the countdown clock and test the claim experience in production.

PR #186, opened 2022-07-08
Timer polish / Admin tools

sparrowDom:
- minor fixes so that claim page re-renders when timer reaches 0
- homepage redirects to claims by default (we should change that when we enable proposal creation)
- Adds Admin tools to be able to override claim open/closes timers (see screenshot below). Menu is enabled by running `localStorage.setItem("admin", "true");` from the browser console

Screenshot 2022-07-08 at 14 02 54
 
**fixes:** https://github.com/OriginProtocol/ousd-governance/issues/171 https://github.com/OriginProtocol/ousd-governance/issues/184

### DIFF
--- a/client/components/holding/Page.tsx
+++ b/client/components/holding/Page.tsx
@@ -7,6 +7,7 @@ import Card from "../Card";
 import CardDescription from "../CardDescription";
 import CardStat from "../CardStat";
 import ExternalLinkIcon from "../ExternalLinkIcon";
+import { useStore } from "utils/store";
 
 const renderer: CountdownRendererFn = ({ days, hours, minutes, seconds }) => (
   <CardGroup horizontal fourCol dontStackOnMobile>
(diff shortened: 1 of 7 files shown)


client/components/holding/Page.tsx
```diff
@@ -29,86 +30,89 @@ const renderer: CountdownRendererFn = ({ days, hours, minutes, seconds }) => (
   </CardGroup>
 );
 
-let date = new Date(0);
-date.setUTCSeconds(parseInt(process.env.CLAIM_OPENS || "0"));
+const HoldingPage = () => {
+  const { claim } = useStore();
+  let date = new Date(0);
+  date.setUTCSeconds(parseInt(claim.claimOpensTs || "0"));
```
sparrowDom, line 37: haven't changed anything in this file below this comment...

client/utils/index.tsx
```diff
@@ -123,23 +123,23 @@ const fetcher = (url: string) => fetch(url).then((res) => res.json());
...
+  return parseInt(useStore.getState().claim.claimClosesTs || "0");
 }
 
-const now = Math.floor(new Date().getTime() / 1000);
+const now = () => Math.floor(new Date().getTime() / 1000);
 
 export function claimOpenTimestampPassed() {
   if (!claimOpensTimestamp()) return true;
-  return now > claimOpensTimestamp();
+  return now() > claimOpensTimestamp();
```
sparrowDom, line 137: This was needed otherwise a "cached" now time was used.

tomhirst (commented):
This is super cool.

LGTM except for 1 small nitpick. We get a console error on the initial load. It looks like the server-side content is mismatching with the client-side, which Next.js doesn't like. I think it's the timer seconds:

Screenshot 2022-07-08 at 13 35 04

client/utils/index.tsx
tomhirst: Smart.

sparrowDom:
tnx @tomhirst... fixed with useEffect and timeout, instead of passing date object from the page down to the component (if server side props would be used)... annoying code messy workaround -.-

PR #197, opened 2022-07-08
Lockups DB fix

tomhirst:
This addresses a bug that meant lockups weren't being added correctly to the database.

Because we enrich the lockup data by requesting from the contracts on-chain, we don't need to store anything other than `address` and `lockupId` at the time of the event.

This had been changed in some places, but not in the prisma schema.

- Schema updated
- Migration command ran
- Small timestamp fix added to vote escrow lockup creation (sidequest while I was here)


sparrowDom:
LGTM

PR #198, opened 2022-07-08
Contract timings uniformity

tomhirst:
Since we've decided that 1 month = 30 days, we should make sure we have uniformity across the dApp and contracts.

I've updated the dApp earlier, and this brings the contracts in line. I've not created any Brownie tests (I'm not sure if it breaks our current ones), but have tested locally — this will definitely need eyes on to check.

Here's me locking up from both types of claim and from the vote escrow. As you can see the end dates now align, no matter where lockups are created from.

Screenshot 2022-07-08 at 19 14 39

Note: @sparrowDom is investigating a potential bug in the staking rewards calculations (if you notice these numbers being off in the screenshot). [ref #199]

### DIFF
--- a/client/components/AdminUtils.js
+++ b/client/components/AdminUtils.js
@@ -15,7 +15,7 @@ const AdminUtils = () => {
     useStore.setState({
       claim: {
         ...useStore.getState().claim,
-        claimOpensTs
+        claimOpensTs,
       },
     });
   };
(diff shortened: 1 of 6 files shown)


franckc:
We are running out of time so let's not update OgvStaking - we can do that after the launch.

Regarding MandatoryLockupDistributor.sol, I think it's ok to change since we haven't deployed to production.
But I will let @tomlinton make the final call.

tomlinton:
Yea this looks fine to me for the mandatory side, it does break the tests. I'll fix them up shortly if no one beats me.

tomlinton:
Fixed the tests and reverted the changes to OgvStaking which we can do after launch.

PR #199, opened 2022-07-08
fix preview points calculation

sparrowDom:
~~I am not 100% sure what the role of the `epoch` is in the contract. Whether it was suppose to be a rounding period to which all the timestamps "snap" to or something else. Will leave it to @DanielVF to sort it out, or maybe anyone else has an idea?~~

~~So I've made a change where the calculation of preview points is fixed and starts returning "normal" values. There is probably some intended logic in there that we still want to get in place before the launch.~~

### DIFF
--- a/contracts/OgvStaking.sol
+++ b/contracts/OgvStaking.sol
@@ -216,10 +216,8 @@ contract OgvStaking is ERC20Votes {
     {
         require(duration >= minStakeDuration, "Staking: Too short");
         require(duration <= 1461 days, "Staking: Too long");
-        //uint256 start = block.timestamp > epoch ? block.timestamp : epoch;
-        uint256 start = block.timestamp;
+        uint256 start = block.timestamp > epoch ? block.timestamp : epoch;
         uint256 end = start + duration;
-        //uint256 endYearpoc = ((end - epoch) * 1e18) / 365 days;
         uint256 endYearpoc = (duration * 1e18) / 365 days;
         uint256 multiplier = PRBMathUD60x18.pow(YEAR_BASE, endYearpoc);
         return ((amount * multiplier) / 1e18, end);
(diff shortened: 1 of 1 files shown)


contracts/OgvStaking.sol
```diff
@@ -216,9 +216,11 @@ contract OgvStaking is ERC20Votes {
     {
         require(duration >= minStakeDuration, "Staking: Too short");
         require(duration <= 1461 days, "Staking: Too long");
-        uint256 start = block.timestamp > epoch ? block.timestamp : epoch;
```
sparrowDom, line 219: epoch in deploy script: https://github.com/OriginProtocol/ousd-governance/blob/master/scripts/deploy.py#L6 is 86400 (seconds in a day). For that reason above will always default to `block.timestamp`

franckc: On mainnet the epoch is correctly set to the start of the staking which is timestamp 1657584000 (Tuesday, July 12, 2022 12:00:00 AM UTC).

See https://etherscan.io/address/0x0c4576ca1c365868e162554af8e385dc3e7c66d9#readProxyContract.

It is NOT set to 86400

franckc: Regarding the role of the epoch: 

    /// If staking before the start of staking (epoch), then the lockup start
    /// and end dates are shifted forward so that the lockup starts at the
    /// epoch.

https://github.com/OriginProtocol/ousd-governance/blob/f3b91ffbe7f54414418e42148420978235502d85/contracts/OgvStaking.sol#L93

sparrowDom: Got it thanks

sparrowDom:
actually there is nothing to do here once the correct epoch is deployed to network. @DanielVF sorry for so much fuss on so many channels :)

PR #203, opened 2022-07-08
Exclude accounts from airdrop data

tomlinton:


retroactive/src/ogn.ts
```diff
@@ -370,14 +370,21 @@ const calculateRewards = async () => {
     ),
   ];
 
+  const exclusionList = (fs.readFileSync("./exclusions.csv", "utf-8")).split("\n").map(ethers.utils.getAddress);
```
franckc, line 373: Wonder if it's worth console.log how many addresses were loaded from the exclusion list.

franckc (approved):
Nice! LGTM

PR #205, opened 2022-07-09
Listener: allow start before claiming and fix small bug

franckc:
Maybe I'm missing some context but I can't think of a reason we would not want the listener to be starting off before the claiming period starts. Uncommented that check. It was causing the listener to not come up in production.

Also fixed a bug where when restarting in production, the listener would always start at block 0. Which would cause reprocessing all the events at every restart.


sparrowDom:
I see no reason for this as well. Merging this in...
Feel free to object @tomlinton @tomhirst if there is some reason for this that we don't understand

PR #207, opened 2022-07-09
add rinkeby deployment artifacts

sparrowDom:
- add rinkeby deployment artifacts
- and copies merkle trees from prod to rinkeby


sparrowDom:
merging this in since it only affects Rinkeby. And so staging can be re-deployed without needing to remember to merge this branch in. Happy to revert / address things subsequenltly

PR #208, opened 2022-07-09
add error message when APY 0% because of 0 emissions

sparrowDom:
add error message when APY 0% because of 0 emissions


sparrowDom:
Merging this in... happy to address anything subsequenlty

PR #211, opened 2022-07-09
fix constraint bug

sparrowDom:
Removes lockupId constraint and add unique auto increment id -> requirement of Prisma. Tested locally, entry created in db and UI: 
Screenshot 2022-07-10 at 00 44 22
Screenshot 2022-07-10 at 00 44 45

### DIFF
--- a/client/prisma/schema.prisma
+++ b/client/prisma/schema.prisma
@@ -42,6 +42,7 @@ model Lockup {
   user      String
   createdAt DateTime @default(now()) @map("created_at")
 
+  @@unique([lockupId, user])
   @@index([user], type: BTree)
   @@map("lockups")
 }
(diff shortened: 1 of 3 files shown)


client/prisma/schema.prisma
```diff
@@ -37,7 +37,8 @@ model Listener {
 }
 
 model Lockup {
-  lockupId  Int @unique
+  id        Int @default(autoincrement()) @id
+  lockupId  Int
```
franckc, line 41: I'd be inclined to have a unique constraint on (user, lockupId) so that we don't ever have duplicates (might happen if for some reason we would run 2 listeners for a brief period of time?)

sparrowDom: Good thinking added another migration for unique constraint
Screenshot 2022-07-10 at 00 56 33

franckc (approved):
LGTM

sparrowDom:
thanks for deploying @franckc

Issue #131, opened 2022-07-05
Batch claimable amounts with token sums


micahalcorn:
If we're going to have so many line items on the eligibility page, let's provide totals so that there is some continuity between the amounts shown here and the amounts on the claim/lock screen. Let's conditionally show a total for OGV and veOGV but only if there are multiple line items that the user is eligible for. The majority of users who only get OGV for holding OGN should not have any sums.

![ogv_03](https://user-images.githubusercontent.com/273937/177409663-972f8654-97cb-4670-b84a-f3cdca42966a.png)

micahalcorn:
<img width="652" alt="Screen Shot 2022-07-07 at 10 49 35 PM" src="https://user-images.githubusercontent.com/273937/177925655-2d5d6b8d-54c0-4701-96e5-3b2fc050f501.png">

sparrowDom:
done via: https://github.com/OriginProtocol/ousd-governance/pull/241

PR #214, opened 2022-07-10
Show locked OGV not veOGV in eligibility table

tomhirst:
I just noticed this while testing:

It doesn't make sense to show veOGV as the token received in the eligibility phase for OUSD holders, because the amount and token type don't align.

For instance: If someone qualifies for the airdrop from being an OUSD (or wOUSD) holder, they receive the amount of OGV specified, but pre-locked. They don't receive the amount of OGV 1 for 1 as veOGV. It's actually higher than we show.

This could also cause confusion at the final claim stage where we show locked OGV eligible for then the veOGV to be received today.

Proposed changes to the token table, from:

Screenshot 2022-07-10 at 11 54 13

To:

Screenshot 2022-07-10 at 11 57 40

### DIFF
--- a/client/components/claim/EligibilityItem.tsx
+++ b/client/components/claim/EligibilityItem.tsx
@@ -39,12 +39,12 @@ const EligibilityItem: FunctionComponent<EligibilityItemProps> = ({
           <div className="flex space-x-2 items-center">
             <TokenIcon
               src="/ogv.svg"
-              alt={showOgvToken ? "OGV" : "locked OGV"}
+              alt={showOgvToken ? "OGV" : "pre-locked OGV"}
             />
             <ReactTooltip id={id} place="top" type="dark" effect="solid">
               <span>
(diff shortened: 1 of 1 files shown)


tomhirst:
Screenshot 2022-07-10 at 11 59 17

110 OGV and 587.66 veOGV shown at claim step, not 110 veOGV.

tomhirst:
Update on this, regarding an idea from @sparrowDom 

Screenshot 2022-07-10 at 19 49 22

micahalcorn:
Good call, I think I prefer @tomhirst's "pre-locked OGV" for a few reasons:

1) The user hasn't been introduced to the concept of veOGV yet
2) It simplifies the [sum of how much you get to claim](https://github.com/OriginProtocol/ousd-governance/issues/131) if we wanted to add that
3) It avoids the confusion that I have with our veOGV icon looking like the green checkmark

sparrowDom:
cool, merging in as is then

PR #218, opened 2022-07-10
Update seconds in a month variable in the dApp

tomhirst:
This updates the variable we use across the dApp for seconds in a month.


tomhirst:
Merging this because we need it now.

sparrowDom (commented):
cool LGTM

Issue #230, opened 2022-07-11
Offer OGV to ineligible users


micahalcorn:
On the eligibility step, make trying another address the secondary CTA and focus on buying OGV as the suggested action.

- [x] Move "Try another address" to a small link below the currently-connected address
- [x] Change the button style to a filled/solid green instead of outline/wire
- [x] Change the button text to "Buy OGV"
- [x] Point the button to https://www.huobi.com/exchange/ogv_usdt
- [x] Add larger text between the link and the button saying "OGV stakers earn a 123.45% variable APY"

<img width="407" alt="Screen Shot 2022-07-11 at 1 28 12 AM" src="https://user-images.githubusercontent.com/273937/178221864-ba313146-0ee1-42b4-8067-ce4e773eafa4.png">

PR #237, opened 2022-07-11
Add APY to staking page for users with 0 balance and 0 stakes

tomhirst:
Addresses #230 

Screenshot 2022-07-11 at 12 42 23


sparrowDom (approved):
Works nicely

PR #242, opened 2022-07-11
update video links

sparrowDom:
update video links


sparrowDom:
will merge this in and deploy it. Have tested the claim experience so each video is in the correct section

PR #246, opened 2022-07-11
adjust lockup gas limits

sparrowDom:
adjust lockup gas limits

### DIFF
--- a/client/components/vote-escrow/LockupForm.tsx
+++ b/client/components/vote-escrow/LockupForm.tsx
@@ -242,7 +242,8 @@ const LockupForm: FunctionComponent<LockupFormProps> = ({ existingLockup }) => {
         transaction = await contracts.OgvStaking["stake(uint256,uint256)"](
           ethers.utils.parseUnits(lockupAmount),
           duration,
-          { gasLimit: 296559 }
+          // 228123 * 1.5
+          { gasLimit: 342184 }
         );
       } catch (e) {
         setTransactionError("Error locking up!");
--- a/client/utils/useClaim.tsx
+++ b/client/utils/useClaim.tsx
@@ -127,7 +127,8 @@ const useClaim = () => {
                   claim.optional.amount,
                   claim.optional.proof,
                   duration,
-                  { gasLimit: 361500 }
+                  // 278048 * 1.5
+                  { gasLimit: 417072 }
                 );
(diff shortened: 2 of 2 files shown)


client/components/vote-escrow/LockupForm.tsx
```diff
@@ -302,8 +302,8 @@ const LockupForm: FunctionComponent<LockupFormProps> = ({ existingLockup }) => {
         transaction = await contracts.OgvStaking["extend(uint256,uint256)"](
           existingLockup.lockupId,
           duration,
-          { gasLimit: 1000000 }
-        ); // @TODO maybe set this to lower
+          { gasLimit: 154727 }
```
franckc, line 305: This is tight.
Tom did an extend on mainnet
https://etherscan.io/tx/0x3f0ba0620459079f8b2d55c380f2409aa876e8776c2b481281af09061ebebb3a

It used 149,604 

How about 150k*1.3 = 195k gas?

client/components/vote-escrow/LockupForm.tsx
```diff
@@ -188,8 +188,8 @@ const LockupForm: FunctionComponent<LockupFormProps> = ({ existingLockup }) => {
       transaction = await contracts.OriginDollarGovernance.approve(
         contracts.OgvStaking.address,
         ethers.constants.MaxUint256,
-        { gasLimit: 1000000 }
-      ); // @TODO maybe set this to lower
+        { gasLimit: 66983 }
```
franckc, line 191: That looks too tight. We did an approve from the multisig and it used 149,604
https://etherscan.io/tx/0x3f0ba0620459079f8b2d55c380f2409aa876e8776c2b481281af09061ebebb3a

tomhirst:
We are missing a figure for `unstake` transactions. I guess users won't be able to call this until at least 1 month has passed. We should be OK to leave for now, but we should remember that we need to test and add a more accurate amount.

client/components/vote-escrow/LockupForm.tsx
franckc: Wrong hash. Here is the one:
https://etherscan.io/tx/0x059faaef43ca156e2da0624bf63a04cd3cb841f1da11c6772f70dd0577b38b17
 
Using 111,065

franckc (approved):
LGTM

Issue #249, opened 2022-07-12
Listener: Inserts into the lockup table failing


franckc:
In production, I'm seeing a few errors in the listener logs related to an insert into the lockups table failing due to the unique constraint on (address, lockupId) firing. See below an example for address 0x73f8dC6e117b096600660c1245ae8E396cB6750e

That would happen if the listener would try to insert an lockup already existing... Could it be fired when someone takes an extend lockup action?

```
2022-07-12T01:05:36.000000+00:00 app[postgres.972634]: [DATABASE] [23-1]  sql_error_code = 23505 time_ms = "2022-07-12 01:05:36.381 UTC" pid="114288" proc_start_time="2022-07-12 00:25:29 UTC" session_id="62ccbf79.1be70" vtid="3/36592" tid="16656" log_line="40" database="da9rlgj2g7ouh5" connection_source="54.234.251.233(36498)" user="udds5kdsfaeo27" application_name="[unknown]" ERROR:  duplicate key value violates unique constraint "lockups_lockupId_user_key"
2022-07-12T01:05:36.000000+00:00 app[postgres.972634]: [DATABASE] [23-2]  sql_error_code = 23505 time_ms = "2022-07-12 01:05:36.381 UTC" pid="114288" proc_start_time="2022-07-12 00:25:29 UTC" session_id="62ccbf79.1be70" vtid="3/36592" tid="16656" log_line="41" database="da9rlgj2g7ouh5" connection_source="54.234.251.233(36498)" user="udds5kdsfaeo27" application_name="[unknown]" DETAIL:  Key ("lockupId", "user")=(2, 0x73f8dC6e117b096600660c1245ae8E396cB6750e) already exists.
2022-07-12T01:05:36.000000+00:00 app[postgres.972634]: [DATABASE] [23-3]  sql_error_code = 23505 time_ms = "2022-07-12 01:05:36.381 UTC" pid="114288" proc_start_time="2022-07-12 00:25:29 UTC" session_id="62ccbf79.1be70" vtid="3/36592" tid="16656" log_line="42" database="da9rlgj2g7ouh5" connection_source="54.234.251.233(36498)" user="udds5kdsfaeo27" application_name="[unknown]" STATEMENT:  INSERT INTO "public"."lockups" ("lockupId","user","created_at") VALUES ($1,$2,$3) RETURNING "public"."lockups"."id"
2022-07-12T01:05:36.380016+00:00 app[worker.1]: info: Got confirmed block 15124769
[...]

PR #254, opened 2022-07-14
Listener fixes

sparrowDom:
- fixes a bug when deleting a lockup record https://github.com/OriginProtocol/ousd-governance/issues/249
- adds more field to lockup table
- adds `ACCOUNT_TO_FUND` env variable that funds account with OGV on node startup


sparrowDom:
@tomhirst merging this in to restart the listener on prod and see if it solves the issue

Issue #253, opened 2022-07-14
Show accrued rewards


micahalcorn:
Users expect to see how much additional OGV they have earned once they have staked. We will also want to add a feature for them to claim these rewards soon.

micahalcorn:
One idea is to have a fourth module below the 3 balances. Make it as wide as the container (equal to 3 balances modules) and then include a line going from the Staked module to the new one. In it, we show "Accrued rewards" or "Claimable rewards" and then have a growing, animated number. I'm not sure how predictable the growth of the number is or if we would need to call a function to get the claimable balance intermittently.

<img width="682" alt="Screen Shot 2022-07-14 at 12 14 07 AM" src="https://user-images.githubusercontent.com/273937/178923946-b0e2bbe5-9e1b-4328-87e5-9eb50eb27d81.png">

micahalcorn:
We could also blend this in with [the CTA that we're planning to show to all users](https://github.com/OriginProtocol/ousd-governance/issues/251#issuecomment-1181311738). I could imagine a module communicating the following:

- Here's how much rewards you've accumulated (potentially zero); claim
- Here's the rate at which rewards are currently accumulating; stake +/- buy

micahalcorn:
One other consideration is whether the claim is tied to unstaking or not. My assumption is that rewards accrue based on the entire combined stake position, not individually for each stake (like they did in [the old program](https://ousd.com/earn). Either way, I would expect the common behavior to be "Claim and unstake" rather than claim and separately unstake. The current design implies that each stake must be unstaked separately, so I would be confused if I am claiming rewards across all stakes but have to unstake them individually.

tomhirst:
I like the idea of adding a "Rewards accrued" card to the total balances panel. We can leverage the `previewRewards()` method to get the OGV figure that would be paid out should `collectRewards()` be called.

`collectRewards()` is called every time someone stakes, unstakes or extends. But it's also callable on its own (similar to the scenario you mentioned). 

This raises a couple of questions on the UX side:

1. Anytime someone stakes/unstakes or extends, are we collecting _all_ rewards accrued? If so, we need to make this clear in the UI

2. We have a path that allows users to collect all accrued rewards across all stakes while leaving their stakes untouched. Do we make this the primary CTA of the page with regards to cashing out earned OGV? IE, a "Collect Rewards" button that reflects the withdrawal of the OGV displayed in the account balance

[...]

PR #256, opened 2022-07-14
Pull acrrued rewards figure from staking contract

tomhirst:
Playground to address: #253


tomhirst:
I think the four columns on the account balance stats row works best. I reduced the text and icon sizes slightly to make it fit while still leaving enough room for padding.

I've added some call out text highlighting the % of all OGV locked up and the % vAPY users are receiving. I think this works quite well.

We also have a CTA (and the functionality) to call the `collectRewards()` function should a user have accrued any rewards.  

Here are a few screenshots showing the various states:

1 live stake but 0 rewards to collect
Screenshot 2022-07-18 at 17 47 38

1 live stake with accrued rewards to collect (ignore the 0 accrued rewards balance in the screenshot)
Screenshot 2022-07-18 at 17 47 15

0 live stakes but a non-zero OGV balance
Screenshot 2022-07-18 at 17 48 08

A non-zero OGV balance (ignore all balance figures)
Screenshot 2022-07-18 at 17 49 04

LMK what you think so far @micahalcorn

micahalcorn:
> LMK what you think so far @micahalcorn


Nice, I agree that the four-column header works.

One might argue that the claim rewards function should be located near the rewards balance. I think it's fine for now down at the bottom. Just as a thought exercise, I was wondering what it would be like to have a button below every balance 🤔

|Balance|Staked|Vote Balance|Accrued Rewards|
|:---|:---|:---|:---|
|10,000|20,000|30,000|1,234|
|OGV|OGV|veOGV|OGV|
|~~~~~|~~~~~|~~~~~|~~~~~|
|**Buy**|**Stake**|**Unstake**|**Claim**|

Then you disable the buttons when they aren't actionable. Probably a dumb idea 🤷‍♂️

There's no scenario where you have zero staked and non-zero accrued rewards, correct? Because claiming is all or none and it's included in the unstake?

I think I'm okay with the "Buy OGV" CTA/button only being shown when the OGV balance and veOGV balance are both zero. There's always the nav link if someone wants it. We should pretty up that last screenshot though. I would maybe tone down the text and change it to something like "OGV is available on many of the top exchanges" followed by 3-4 logos and then the button linking to Uniswap.

tomhirst:
> > LMK what you think so far @micahalcorn
> 
> Nice, I agree that the four-column header works.
> 
> One might argue that the claim rewards function should be located near the rewards balance. I think it's fine for now down at the bottom. Just as a thought exercise, I was wondering what it would be like to have a button below every balance 🤔
> 
> Balance	Staked	Vote Balance	Accrued Rewards
> 10,000	20,000	30,000	1,234
> OGV	OGV	veOGV	OGV
> ~~~~~	~~~~~	~~~~~	~~~~~
> **Buy**	**Stake**	**Unstake**	**Claim**
> Then you disable the buttons when they aren't actionable. Probably a dumb idea 🤷‍♂️
> 
> There's no scenario where you have zero staked and non-zero accrued rewards, correct? Because claiming is all or none and it's included in the unstake?
> 
> I think I'm okay with the "Buy OGV" CTA/button only being shown when the OGV balance and veOGV balance are both zero. There's always the nav link if someone wants it. We should pretty up that last screenshot though. I would maybe tone down the text and change it to something like "OGV is available on many of the top exchanges" followed by 3-4 logos and then the button linking to Uniswap.

I see where you're going with the buttons under each column. It would definitely indicate that the action is tied to the amounts in the cells above. My concern would be that in some cases it would be misleading though. For instance:

- Unstake under the Vote Balance amounts would infer every lockup would be unstaked by clicking that button
- Stake under the Staked OGV would infer that the totalled OGV already staked would be then staked again, not that it would take you to create a new stake

Correct. Unstaking calls collectRewards in an all-or-nothing fashion.

We can change the "You can buy..." message wording and tone down the size/weight. What logos would you like to be displayed above the button linking out to Uni? Assuming Huobi, and...

tomhirst:
@micahalcorn Thanks!

Here's the updated screen, LMK if you're happy with this PR and we'll get it merged.

Screenshot 2022-07-21 at 15 01 03

micahalcorn:
@tomhirst I think we should go ahead with these improvements, thanks. It would be nice to emphasize the APY more because it gets a little lost in the text. But we can continue to iterate after this is live in its current form.

micahalcorn (approved):

PR #258, opened 2022-07-14
Use network ID from env var not wallet

tomhirst:
Addresses: #257

It seems like we were aborting the load of the contracts if no wallet was connected, but we need them to call the APY function whether a wallet is connected or not.

I've re-ordered a few things and it seems to work without any unintended consequences elsewhere.

Would love a second set of eyes though, @sparrowDom


sparrowDom (approved):
LGTM

Issue #260, opened 2022-07-15
Rounding issue with some new stakes


micahalcorn:
Several users per day are having their `stake` transactions reverted and they appear to be attempting to stake a round number that is slightly higher than their balance. Interestingly, the most recent example involved a user who was eventually successful.

🔴 [Failed transactions](https://etherscan.io/txs?a=0x0c4576ca1c365868e162554af8e385dc3e7c66d9&f=1)

franckc:
Example of a failure:
https://etherscan.io/tx/0xef83f3fbc32d8cec891bff333bc919c53178242e56be93c6cb0ea175173666a4

Notice the input data for the stake amount is 0x279b756e7b9bc680000 which is 11690*1E+18
But the the user only has 11,689.9877 OGV in their wallet: https://etherscan.io/address/0xabbedbb6b3627640f62ee4a2a5a67a01e92d032c

micahalcorn:
**Steps to reproduce**

- Claim a non-round number of OGV and do not stake
- Visit /stake/new
- Drag the first slider to the end (max)
- Submit the transaction

https://rinkeby.etherscan.io/tx/0xbd36e2fd92ad250649cc3713ad7c55c9be397195ebe1625c8a00decdcd540d6b

PR #261, opened 2022-07-15
Fix rounding on lockup form

tomhirst:
A quick fix to address #260:

- We identified that non-round number 100% stakes would revert due to rounding
- This is because the input slider requires a string version (or "formatted" version) of the actual BigNumber balance so the UX plays nicely
- This results in, non-round, 100% stakes, being passed as slightly higher to the contract, causing a revert due to the user trying to stake more OGV than they have
- **The fix** in this PR detects whether a user wants to stake 100% of their OGV holdings and bypasses the slider input value in favour of their actual balance of OGV (which is untouched, and thus not rounded)
- We only do this when 100% is selected because we can safely assume that the user has enough OGV to cover the slight rounding on any lower percentages 

Screenshots of the fix in flow:

1. User has a non-round amount of OGV to claim
Screenshot 2022-07-15 at 08 17 35

2. User claims without staking
Screenshot 2022-07-15 at 08 17 52
 
3. User stakes 100% of their balance successfully
Screenshot 2022-07-15 at 08 19 04
Screenshot 2022-07-15 at 08 19 14
Screenshot 2022-07-15 at 08 19 23
Screenshot 2022-07-15 at 08 19 43

### DIFF
--- a/client/components/vote-escrow/LockupForm.tsx
+++ b/client/components/vote-escrow/LockupForm.tsx
@@ -10,7 +10,6 @@ import RangeInput from "components/RangeInput";
 import useLockups from "utils/useLockups";
 import useTotalBalances from "utils/useTotalBalances";
 import LockupTable from "components/vote-escrow/LockupTable";
-import moment from "moment";
 import { SECONDS_IN_A_MONTH } from "../../constants/index";
 
 interface LockupFormProps {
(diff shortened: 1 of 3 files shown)


micahalcorn (approved):

sparrowDom (approved):
LGTM

PR #263, opened 2022-07-19
Fix for skipped gas optimization

DanielVF:
There was comparison backwards in RewardsSource.sol, which skipped a gas optimization. This did not change the behavior of the contract, but did result in an extra trip through the loop and an extra storage load.

## Background:

_calcRewards determine how the amount of OGV that is created as inflation. To do this, it checks the range of time that has passed to the current time since the last inflation, and then creates the needed inflation for that time range. 

The function loops through configurable inflation slope time segments, determine how much of the range is in each, and thus the inflation earned in that slope.

image

To speed things up there are two gas optimizations made: 

1. We start our loop at the last used inflation slope, allowing us to skip all the inflation slopes before.
2. If the end of the range fits inside the current slope, we know that no further sloop could overlap, and skip further checks.

Most of the time the range we are working with fits entirely inside the current slope. Because of the two optimizations  the only slope configuration that we need to read from storage is the current slope.

image

## Bug

The comparison check for the second optimization was backwards, checking that the local range end was into the next slope, which would never happen because the local range is trimmed to the slope end, and thus the optimization could never fire.

## Fix

Fixing the bug improves gas usage.

image

Even more importantly, there's not a logic bug sneaking around in the contract, waiting to explode some future change.

Given the small change in gas usage, we won't deploy this fix separately, but it will be a part of any future upgrade.

# Thanks!

Thanks to [okkothejawa](https://twitter.com/okkothejawa) for spotting this!

### DIFF
--- a/contracts/RewardsSource.sol
+++ b/contracts/RewardsSource.sol
@@ -39,7 +39,7 @@ contract RewardsSource is Governable {
     /// @return rewards OGV collected
     function collectRewards() external returns (uint256) {
         require(msg.sender == rewardsTarget, "Rewards: Not rewardsTarget");
-        require(lastRewardTime > 0, "Rewards: lastRewardTime is zero"); // Ensures initialziation
+        require(lastRewardTime > 0, "Rewards: lastRewardTime is zero"); // Ensures initialization
         if (block.timestamp <= lastRewardTime) {
             return 0;
         }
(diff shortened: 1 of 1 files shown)


mikeshultz (approved):
No extra SLOAD, nice.  :+1: 

Might be good to add some tests getting into all these branches?  Non-blocking, LGTM

contracts/RewardsSource.sol
```diff
@@ -99,7 +99,7 @@ contract RewardsSource is Governable {
             if (i > _currentSlopeIndex && duration > 0) {
                 nextSlopeIndex = i; // We have moved into a new slope
             }
-            if (slopeEnd < rangeEnd) {
+            if (rangeEnd < slopeEnd) {
```
DanielVF, line 102: Rather than flipping the sign, I flipped the arguments so that they match the same ordering used through the rest of the function. If I'd done this to start with, the bug probably would have been obvious.

franckc (approved):
LGTM

DanielVF:
Tests were run.

Issue #277, opened 2022-07-26
Emphasize APY on staking screen


micahalcorn:
I think we should further emphasize the APY and potentially also the percentage staked in this message. The numbers get a little lost for me since they're in-line with other text and the contrast between white and green is relatively low.

_Just for demonstration..._

<img width="664" alt="Screen Shot 2022-07-25 at 10 58 15 PM" src="https://user-images.githubusercontent.com/273937/180934484-d361aae8-7423-4243-a23d-c4e321682058.png">
<img width="663" alt="Screen Shot 2022-07-25 at 10 58 28 PM" src="https://user-images.githubusercontent.com/273937/180934490-a8e3a053-1b5b-4fcc-8c86-96ead1f4ab38.png">
<img width="666" alt="Screen Shot 2022-07-25 at 10 59 31 PM" src="https://user-images.githubusercontent.com/273937/180934495-a04bceac-0855-40ab-a7da-7797ddbe7547.png">

PR #283, opened 2022-07-27
Pick out % amounts

tomhirst:
Addresses: #277

Screenshot 2022-07-27 at 11 05 28

@micahalcorn This seems to work well in our style. The darker green we use for positive enforcement of APY percentages elsewhere so it makes sense to stick to the same pattern. Setting the text on two lines helps with the mobile layout and further emphasizes the two numbers in their own right. LMK what you think.


micahalcorn (approved):
I like it, thanks.

Issue #271, opened 2022-07-23
Stake extension slider quirks


micahalcorn:
I think we should allow the slider to go back to zero. A lesser issue is that the cursor can easily lose focus on the slider if the user is too aggressive with the drag.

https://www.loom.com/share/1a7eb00b4c6346c09f45e17e7988f42c

PR #287, opened 2022-08-01
Remove check that prevents slider going below current duration

tomhirst:
Addresses: #271 

- Removes the slider check that prevents stake extensions from being set lower than their existing amount
- Check still exists on submit transaction button
- This also fixes the bug demonstrated in the video from #271 

Screenshot 2022-08-01 at 11 38 35
Screenshot 2022-08-01 at 11 38 29


micahalcorn:
This sounds good to me in principle. I'll let @sparrowDom confirm that there aren't any side effects.

sparrowDom (approved):
nice LGTM

PR #288, opened 2022-08-01
read contract abi from the fragments section

sparrowDom:
- fix how contract abi is being read
- use a general rpc provider when listening for transaction status
- fix bug with reading voting power
- fix favicon errors


tomhirst (approved):
LGTM

Issue #273, opened 2022-07-23
Etherscan links for stake transactions


micahalcorn:
I was expecting to be able to drill down to Etherscan on my existing stakes. This gets a little messy for stakes that I extend because they're not 1:1. The same is also true for veOGV airdrop claims, which are like 4:1. I think it might make sense to have an expandable detail row for each blockchain transaction associated with each stake. So some stakes might have several associated transactions if they've been extended, and some stakes might each have the same associated transaction linked (redundantly) if they were all part of one claim.

tomhirst:
Yep, it's easy for this to become confusing because we have several transactions related to a single staking (or claiming) action. I like your proposal to have a drop-down expander that within lists all related transactions + links to Etherscan.

We are storing stakes in the database anyway, so what we could do, is save an array of transaction addresses to each stake. 

A couple of potential technical thoughts:

1. When the `extend` contract method is called it unstakes the existing stake and then creates a new one. The event listener is creating/deleting stakes in the database by listening to the "Stake" and "Unstake" events to completely remove and re-add user-specific stakes on extension. If we want to retain an ongoing array of transaction addresses, we won't be able to delete a stake and re-add it as we do now. We'll need to re-factor the listener. This might not be straightforward given the event ordering of the contract method. 

2. We would need to run the listener back from the block that the staking contract was created to back-fill the database with transaction data from previous staking and unstaking events. Users may see some oddities in the staking section of the dApp during this time.

PR #290, opened 2022-08-01
Etherscan links for stake transactions

tomhirst:
Addresses: #273 

- Adds an Etherscan icon next to each stake
- Clicking the icon opens up a modal window 
- Modal window offers user detail about each transaction associated with the stake
- With a link out to the Etherscan transaction hash

Screenshot 2022-08-02 at 14 28 37

Screenshot 2022-08-02 at 14 28 44

Note: This refactors the database schema and listener. Once reviewed, we can test on Rinkeby by dropping the `listener` and `lockups` table, setting the start block to the block the contract was deployed, and letting the listener back-fill transaction data within the database.


micahalcorn:
> Addresses: #273

Cute 👌

client/prisma/schema.prisma
```diff
@@ -4,48 +4,60 @@ datasource db {
...
 
   @@unique([lockupId, user])
   @@index([user], type: BTree)
   @@map("lockups")
 }
+
+model Transaction {
+  hash      String   @unique
+  event     String
+  createdAt DateTime @default(now())
```
franckc, line 58: Suggestion: the use of camel case for column names means quotes are required when writing queries on the sql console. Which is a bit of a pain in the neck :)

Moving forward, in the governance repo can we take the convention to have all column names lower case with underscore? That's what we use in all other projects.

I'm not familiar much with prisma but I think just adding `@map("created_at")` might do the trick?  The column name in the DB will be `created_at` while the model defines the attribute as createdAt

Same comment for Transaction.lockupId below.

client/components/vote-escrow/LockupsTable.tsx
```diff
@@ -0,0 +1,166 @@
...
+                <tr key={transaction.hash}>
+                  <td>
+                    {moment(transaction.createdAt).format(
+                      "MMM D, YYYY, HH:mm:ss"
+                    )}
+                  </td>
+                  <td>{transaction.event}</td>
+                  <td>
+                    <Link
+                      href={
```
sparrowDom, line 144: nit: would probably be better for these to open in a new window

client/prisma/migrations/20220802122130_add_transaction_detail/migration.sql
```diff
@@ -0,0 +1,50 @@
...
+DROP INDEX "proposals_proposal_id_key";
+
+-- DropIndex
+DROP INDEX "voters_address_key";
+
+-- AlterTable
+ALTER TABLE "lockups" DROP COLUMN "txHashes";
+
+-- CreateTable
+CREATE TABLE "Transaction" (
```
sparrowDom, line 27: This migration file is overriden by the next one right? If you want I think you can clean it up by force deleting some already applied migrations in `_prisma_migrations` psql table in case you would want to get rid of this file

client/prisma/schema.prisma
sparrowDom: the pains of sql and case insensitivity :)

sparrowDom (commented):
few minor comments otherwise feel free to merge. This is great stuff!

sparrowDom:
oh also deploy wise - we should probably restart the listener to pick up the old transactions?

client/prisma/schema.prisma
```diff
@@ -4,48 +4,60 @@ datasource db {
...
+  Lockup    Lockup?  @relation(fields: [lockupId], references: [id])
+  lockupId  Int
+
+  @@map("transaction")
```
sparrowDom, line 62: nit: to be consistent transaction**s**

client/prisma/schema.prisma
tomhirst: Yeah, this makes total sense. I think as long as the snake case`@maps()` are included, it's not so bad, but moving forward let's abandon camel case column names when using SQL.

client/prisma/migrations/20220802122130_add_transaction_detail/migration.sql
tomhirst: Good call, will delete as we don't need a deep history in dev.

tomhirst:
> oh also deploy wise - we should probably restart the listener to pick up the old transactions?

Yes, was thinking to drop the Rinkeby database tables and restart the listener from the block the contract was deployed.

client/components/vote-escrow/LockupsTable.tsx
tomhirst: > nit: would probably be better for these to open in a new window

Good catch, I have a prop for that ;)

PR #297, opened 2022-08-02
Enable fee distribution to OGV stakers

zskolnik:
Please see https://github.com/OriginProtocol/origin-dollar/pull/1058 (this feature requires changes in both repositories)


franckc:
Thanks for making this change! Can we add a unit test to make sure this works as expected?

contracts/RewardsSource.sol
```diff
@@ -103,6 +104,11 @@ contract RewardsSource is Governable {
                 break; // No future slope could match
             }
         }
+
+        // When previewing or sending rewards, check it's own OGV balance,
+        // and if present, send that along as part of the rewards
+        total += Mintable(ogv).balanceOf(address(this));
```
DanielVF, line 110: This method of handling buyback funds is wrong. It's inside a method that calculates the inflation.

This means that every time inflation is done (every stake / unstake / collect), the inflation is too much by the sum of all buyback funds ever sent to the rewards contract. The OGV sent by the buyback contract never leaves the rewards contract, instead just staying to inflate the mints.

DanielVF (changes requested):
Critical bug. Will need to be fixed. Have updated the test so that the test correctly notices the error.

contracts/RewardsSource.sol
shahthepro: Ah, yes. Sorry I totally missed the `collectRewards()` method. Have updated the code. Thanks for catching this

DanielVF:
@shahthepro I think the "calcRewards()" internal method name may be a little confusing here, perhaps it should be changed to "calculateInflation()" and then the rest of code can get much clearer.

The total rewards should be both the inflation rewards and the balance rewards. We don't pay inflation rewards out of the balance on the contract. So the actual rewards payout is then two independent if's. If we have balance, we pay that. Completely separately, if we don't any inflation, then we we mint the inflation.

shahthepro:
> I think the "calcRewards()" internal method name may be a little confusing here, perhaps it should be changed to "calculateInflation()" and then the rest of code can get much clearer.

> The actual rewards payout is then two independent if's. If we have balance, we pay that. If we have inflation, then we we mint the inflation. We don't need to do two balance checks, once to add in _calcRewards and once to subtract in collectRewards.

@DanielVF Have changed the method name `_calculateInflation()` and have also removed the balance check from this method. However, I had to add the balance check to `previewRewards()` method, which seems to be used by other contracts (`OgvStaking.sol`) and the DApp

DanielVF:
We'll need to make sure slither checks pass before merging to master. In this case is concerned that we are ignoring the final amount from the swap. We don't actually need it, since the check on min size is handled by the swap, so we can safely ignore it.

I think the right comment is something like `// slither-disable-next-line slither-disable-next-line`

DanielVF:
We'll need a deploy test for this PR.

Would also be good to paste the fork tests into the issue so others can play with them.

shahthepro:
Have added the deploy script. Copying my own message from Discord:

- Tried using the map.json file under build/deployments but it didn't have addresses of all deployed contracts. So have just hardcoded the addresses of OGV and RewardsSourceProxy contracts in the code
- Looks like a Multi-sig wallet is the governor of RewardsSourceProxy on mainnet, so there's no governance proposal for this one (let me know if I missed something here)

Also, here's the script I used to test the deployment on fork
```
import os
from brownie import *

def main():
	deployer = accounts.add(os.environ["DEPLOYER_KEY"])
	FROM = {'from': deployer}
	OGV_PROXY_ADDRESS = "0x9c354503c38481a7a7a51629142963f98ecc12d0"
	rewards = run("deploy_002_upgrade_rewards_source")

	ogv = OriginDollarGovernance.at(OGV_PROXY_ADDRESS)
	governor = accounts.at(rewards.governor(), force=True)

	# Allow deployer to mint
	ogv.grantMinterRole(deployer.address,{'from':governor})

	# Fetch current balance and rewards
	(curr_balance, curr_rewards) = fetch_balance(ogv, rewards)

	# Send some OGV to the new contract
	print("Minting 10k OGV to contract")
	ogv_to_mint = 1e4 * 1e18
	ogv.mint(rewards.address, ogv_to_mint, FROM)

	# Fetch new balance
	(new_balance, new_rewards) = fetch_balance(ogv, rewards)

	# Check if any balance added to the contract 
	# is being calculated in the rewards
	print("Expected Balance:", curr_balance + ogv_to_mint)
	print("Actual Balance:  ", new_balance)

	print("Expected Rewards:", curr_rewards + ogv_to_mint)
	print("Actual Rewards:  ", new_rewards)

def fetch_balance(ogv, rewards):
	curr_balance = ogv.balanceOf(rewards.address)
	curr_rewards = rewards.previewRewards()

	print("Current Balance:", curr_balance)
	print("Current Rewards:", curr_rewards)

	return (curr_balance, curr_rewards)
```

DanielVF:
I'll have to give this a full review another day, but it's looking good so far. Thanks!

DanielVF:
### Invariants

- [ ] Critical: The amount of OGV given to the rewards contract (from inflation+other) must be equal to number returned by `collectRewards()`.
- [ ] The amount returned from previewRewards() should match the amount returned from an immediate `collectRewards()`;

DanielVF (approved):

sparrowDom (approved):
LGTM.

We don't need to use SafeMath because of solidity version being > 0.8 right. Have updated the [security checklist's](https://github.com/OriginProtocol/security/blob/master/templates/Contract-Code-Review-Example.md) overflow item: 
Never use "+" or "-", always use safe math or have contract compile in solidity version > 0.8

DanielVF:
Correct, solidity >= 0.8 means no safe math needed.

PR #302, opened 2022-08-03
DB migration fix

tomhirst:
Fixes Prisma migration issue experienced on staging:

- Removes unneeded migration
- Gives active a default value

Note:

The database tables on https://ousd-governance-rinkeby.herokuapp.com/ were also dropped to allow this to work. It is safe to do this because the listener backfills also content as it would have been created in real-time.

The listener table's last_seen_block value was also set to the block the staking contract was deployed at.

When we next merge master to stable we should remember this database work and take a backup pre-drop.


sparrowDom (approved):
LGTM

Issue #304, opened 2022-08-04
Improve disconnected state for staking


micahalcorn:
Not sure how I just now realized this, but the default path for many users at /stake is not exactly compelling:

<img width="654" alt="Screen Shot 2022-08-03 at 9 02 14 PM" src="https://user-images.githubusercontent.com/273937/182760518-c8f274b8-7c28-45c5-aaaa-a18706b51dfd.png">

Much like https://github.com/OriginProtocol/origin-dollar/issues/1007, we should allow the user to have as much of the experience as possible prior to connecting a wallet. The user should approximately see the same elements as she would if a wallet were connected (particularly the staking percentage and APY):

<img width="665" alt="Screen Shot 2022-08-03 at 9 05 11 PM" src="https://user-images.githubusercontent.com/273937/182760690-be503775-32d9-44a8-9746-de13b6130022.png">

For the numbers at the top, we could treat them like we do for disconnected users on the swap screen:

<img width="719" alt="Screen Shot 2022-08-03 at 9 07 15 PM" src="https://user-images.githubusercontent.com/273937/182760892-988466ee-a689-44a8-bec5-dd1201d23a65.png">

Then, below the APY, it probably makes sense to keep the logos intact but replace the button with a "Connect wallet" button.

franckc:
@tomhirst Would you be ok taking this on during this sprint? ty sir...

PR #305, opened 2022-08-04
Improve disconnected state for staking

tomhirst:
Addresses: #304 

Screenshot 2022-08-04 at 13 14 33

### DIFF
--- a/client/components/TokenAmount.tsx
+++ b/client/components/TokenAmount.tsx
@@ -8,11 +8,7 @@ interface TokenAmount {
   format?: string;
 }
 
-const TokenAmount: FunctionComponent<TokenAmount> = ({
-  amount,
-  format,
-  disconnected,
-}) => {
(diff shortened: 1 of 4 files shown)


client/components/TokenAmount.tsx
```diff
@@ -1,13 +1,20 @@
...
 interface TokenAmount {
   amount: BigNumber | string | number;
   format?: string;
 }
 
-const TokenAmount: FunctionComponent<TokenAmount> = ({ amount, format }) => {
+const TokenAmount: FunctionComponent<TokenAmount> = ({
+  amount,
+  format,
+  disconnected,
```
sparrowDom, line 14: this isn't used anywhere right?

sparrowDom (approved):
one nitpick inline otherwise LGTM

client/components/TokenAmount.tsx
tomhirst: Good catch ser, removed!

PR #320, opened 2022-08-12
Governance v1

tomhirst:
This PR gets the governance part of the dApp up to a usable v1 level:

- Ensures dApp interacts with the governance smart contracts properly
- Fixes database issues and enriches data with connections
- Borrows some UI from https://compound.finance/governance

Once tested as functional on staging + merged, we can look at adding improvements (with product + design input) in smaller PRs.

**Don't merge prior to testing on Rinkeby**

Some screenshots from my local environment:

Screenshot 2022-08-12 at 14 50 19

Screenshot 2022-08-12 at 14 50 29

Screenshot 2022-08-12 at 14 50 38

Screenshot 2022-08-12 at 14 50 45

Screenshot 2022-08-12 at 14 51 02


DanielVF:
Does this work okay if there is an ampersand or gt or less than in the title? I'm wondering if this is going to be double escaping things.

tomhirst:
> Does this work okay if there is an ampersand or > or < than in the title / description? I'm wondering if this is going to be double escaping things.

It does now. Good catch, ser.

Screenshot 2022-08-12 at 15 02 43

Screenshot 2022-08-12 at 15 06 02

DanielVF:
Wearing my paranoid hat here, since an escaping attack on the voting UI would be critical.

I'm concerned that since the `dangerouslySetInnerHTML`'s are in a separate file from the actual escaping, that future changes could inadvertently mess things up.

In `getProposalContent` can we add a comment that the output of this function must be escaped and safe for use as raw HTML in the UI? And could we change the names of in output to append either "safe", "sanitized" or "clean", so that when writing the front end code you know that it's been checked elsewhere.

Ideally we'd have unit tests checking that no scripts could make it through, but we can skip this since we don't have front end browser tests.

tomhirst:
> Wearing my paranoid hat here, since an escaping attack on the voting UI would be critical.
> 
> I'm concerned that since the `dangerouslySetInnerHTML`'s are in a separate file from the actual escaping, that future changes could inadvertently mess things up.
> 
> In `getProposalContent` can we add a comment that the output of this function must be escaped and safe for use as raw HTML in the UI? And could we change the names of in output to append either "safe", "sanitized" or "clean", so that when writing the front end code you know that it's been checked elsewhere.
> 
> Ideally we'd have unit tests checking that no scripts could make it through, but we can skip this since we don't have front end browser tests.

Good points. Suggestions added.

DanielVF (approved):

tomhirst:
Updates from recent feedback post-demo:

- The creation of new proposals from the UI has been disabled
- Truncated addresses now link out to Etherscan

Screenshot 2022-08-17 at 11 56 32

matt-bullock:
Will the active tag update to executed when the proposal is completed? It might be good to show a proposal history on the proposal pages.

Could add a back button back to all proposals. I'm used to seeing these on governance UIs.

Could visualise the vote allocation with some kind of bar chart/line diagram like so:
image

I'm unable to click on the contract address in the description box.

I'm not really sure what Arguments and Argument types are. Tbh the whole "Description" box doesn't make much sense to me.
image

Same for the details box. We could look at simplifying this to make it more understandable.

Small UI thing but I struggle to read the text on these disabled buttons. Could up the contrast in the button text so its readable enough whilst still looking disabled. 
image

tomhirst:
> Will the active tag update to executed when the proposal is completed? It might be good to show a proposal history on the proposal pages.
> 
> Could add a back button back to all proposals. I'm used to seeing these on governance UIs.
> 
> Could visualise the vote allocation with some kind of bar chart/line diagram like so: image
> 
> I'm unable to click on the contract address in the description box.
> 
> I'm not really sure what Arguments and Argument types are. Tbh the whole "Description" box doesn't make much sense to me. image
> 
> Same for the details box. We could look at simplifying this to make it more understandable.
> 
> Small UI thing but I struggle to read the text on these disabled buttons. Could up the contrast in the button text so its readable enough whilst still looking disabled. image

Great feedback, thanks @matt-bullock.

1. Yep, the tags update depending on the state of the proposal. Good shout on the history, I've made the relevant changes to the listener to monitor each proposal's state and save timestamps each time it changes.
2. I've added a back button at the top of the single proposals template.
3. Went with the Compound-like progress bars for For/Against voting visualisation.
4. Fixed the contract address not linking out to Etherscan.
5. The description box's purpose is to show two things. 1, the brief description of the proposal submitted on-chain by the user (some of the proposals on RInkeby have no description, which is a little confusing) and 2, to show the actions that will be taken on-chain should the proposal succeed. I'm in agreement that it's not totally clear, though I'm not sure how to improve this visually.
6. Re: details box. We potentially drop this now that we have the history section. Again, I'm not sure how we'd simplify it visually and I'm concerned some users might want to see this information.
7. Good catch on the disabled button text being hard to read. I've increased the contrast there.

From a technical point of view, everything Compound has on screen, we now have. It doesn't look as good as what they have though. Should we invest some time rejigging things in Figma to attain a better layout? Let me know what you think @matt-bullock and @micahalcorn.

tomhirst:
Screenshot 2022-08-19 at 15 09 13

Screenshot 2022-08-19 at 15 09 20

micahalcorn:
Video overload 🌊

**P2**
- [ ] [Proposals UX and level of detail](https://www.loom.com/share/1ec4c4e3e6d2426c9e096ae67b13ead7)
- [ ] [Proposal detail confusion](https://www.loom.com/share/37aa4fbcec38426680873d38317c2cc6)
- [ ] [Progress confusion](https://www.loom.com/share/616dde65a26c4f6999daffb10764e5c5)

**P3**
- [ ] [Improve proposal states](https://www.loom.com/share/ef401c790ee34b98bca30ec51aee05ad)
- [ ] [Avoid truncating addresses](https://www.loom.com/share/44790243d2354165b6ac42b29d0493f7)
- [ ] [Support new proposals in test environment](https://www.loom.com/share/36375a8c61734fef8c46a1de9415c103)
- [ ] [Unnecessary Etherscan logos](https://www.loom.com/share/0e82bb2164114380ae9ac2b0fe9c70bd)

tomhirst:
> @tomhirst @micahalcorn I'd be happy to work on the design for this. As we have a new brand style/ look and feel which we will soon be introducing, would it would make sense for me to start re-designing it in this style?

Nice! Good point, though. Perhaps my v1 will be good enough from a design perspective until we roll out the wider re-style across Origin products? Though if you can think of any quick layout wins in the current style, let me know and I'll implement them.

matt-bullock:
I think this is fine for v1. I look forward to giving it a proper redesign when the time comes. A couple of quick wins that would help in the meantime:

Being able to click the entire area of the proposal to go into it, rather than specific text links. Hopefully this makes sense:

image

---

It would be good to show the for and against percentage bars even if one side has no votes (it would just be the empty grey bar). Like you have here:

image


---

Small alignment thing. You can line the text up here so it's all aligned 

image

---

For the progress bar in the details section it would help to have the grey bar in the background so show where the current progress sits in relations to being 100% complete. Like you have the grey background in the for and against vote bars.

image

tomhirst:
> Video overload 🌊
> 
> **P2**
> 
> * [x]  [Proposals UX and level of detail](https://www.loom.com/share/1ec4c4e3e6d2426c9e096ae67b13ead7)
> * [x]  [Proposal detail confusion](https://www.loom.com/share/37aa4fbcec38426680873d38317c2cc6)
> * [x]  [Progress confusion](https://www.loom.com/share/616dde65a26c4f6999daffb10764e5c5)
> 
> **P3**
> 
> * [x]  [Improve proposal states](https://www.loom.com/share/ef401c790ee34b98bca30ec51aee05ad)
> * [ ]  [Avoid truncating addresses](https://www.loom.com/share/44790243d2354165b6ac42b29d0493f7)
> * [x]  [Support new proposals in test environment](https://www.loom.com/share/36375a8c61734fef8c46a1de9415c103)
> * [x]  [Unnecessary Etherscan logos](https://www.loom.com/share/0e82bb2164114380ae9ac2b0fe9c70bd)

Thanks for this @micahalcorn - I've worked through them all and have made some additions to the staging environment.

tomhirst:
@micahalcorn Here are a few screenshots from my local testing to add a little colour to the different proposal states:

**Proposal index**

Screenshot 2022-08-25 at 13 15 32

**Pending proposal**

Screenshot 2022-08-25 at 13 16 13

**Active proposal**

Screenshot 2022-08-25 at 13 16 36

**Defeated proposal**

Screenshot 2022-08-25 at 13 17 01

**Succeeded proposal**
Screenshot 2022-08-25 at 13 17 22

Note: In a non-local environment, proposals that succeed will eventually be queued and executed. These events will be picked up by the listener and will further populate the proposal history panel with timestamps and transactions.

micahalcorn:
> Note: In a non-local environment, proposals that succeed will eventually be queued and executed. These events will be picked up by the listener and will further populate the proposal history panel with timestamps and transactions.

Amazing, thanks 👏 I'd suggest that we go ahead and proceed. We can continue to test with the new staging contracts and log potential improvements while we move forward as-is with production.

client/components/proposal/SupportTable.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <tbody>
+        {voters.map((voter) => (
+          <tr key={voter.address}>
+            <td className="pl-0">
+              <Address address={voter.address} />
+            </td>
+            <td>
+              <TokenAmount
+                amount={ethers.utils.formatUnits(
+                  ethers.BigNumber.from(voter.votes)
```
shahthepro, line 28: nit: BigNumber.from can throw if voter.votes isn't BigNumberish

client/components/proposal/ProposalVoteStats.tsx
```diff
@@ -1,78 +1,106 @@
...
 
 export const ProposalVoteStats = ({
   proposal,
   votePower,
   onVote,
   hasVoted,
+  votingActive,
 }) => {
+  let tooltipText = "";
+  if (!votePower.gt(0)) {
```
shahthepro, line 17: Would be better to check if it's a BigNumber before checking the value

client/components/proposal/ProposalTable.tsx
```diff
@@ -30,58 +37,51 @@ export const ProposalTable = ({ proposalData }) => {
...
-              <td>
-                <StateTag state={proposalData.states[index]} />
-              </td>
-            </tr>
-          ))}
+          {proposalData?.proposals.map((proposal, index) => {
+            const { cleanTitle } = getCleanProposalContent(
+              proposal.description
+            );
+            const id = proposal?.displayId.toString().padStart(3, "0");
```
shahthepro, line 52: Any reason why proposal var would be null/undefined? The check is missing in the line above but exists here and on the next one. 

Would it be better to not render the row if it happens to be undefined?

client/components/proposal/ProposalTable.tsx
```diff
@@ -30,58 +37,51 @@ export const ProposalTable = ({ proposalData }) => {
...
+            const lastTx =
+              proposal?.transactions[proposal?.transactions.length - 1];
+
+            return (
+              <tr
+                key={index}
+                className="cursor-pointer hover:bg-gray-100"
+                onClick={() => router.push(`/proposals/${proposal[0]}`)}
```
shahthepro, line 60: proposal is treated as an object above and as an array here, guess it might be a bug

client/components/proposal/ProposalDetail.tsx
```diff
@@ -95,42 +156,84 @@ export const ProposalDetail = ({
...
+        <div className="mr-4 relative max-w-lg">
+          <PageTitle noBottomMargin>
+            <div
+              className="break-words"
+              dangerouslySetInnerHTML={{ __html: cleanTitle }}
+            />
+          </PageTitle>
+          <div className="text-white opacity-70 text-md">
+            {id.toString().padStart(3, "0")} •{" "}
+            {`${lastTx?.event} ${moment(lastTx?.createdAt).format(
```
shahthepro, line 173: If lastTx is null, this is gonna show as undefined

client/components/proposal/ProposalActionsTable.tsx
```diff
@@ -21,14 +21,14 @@ export const ProposalActionsTable = ({
...
     "115792089237316195423570985008687907853269984665640564039457584007913129639935";
 
-  const { web3Provider, contracts } = useStore();
+  const { rpcProvider, contracts } = useStore();
   const [modalOpen, setModalOpen] = useState(false);
   const [actionDeleteIndex, setActionDeleteIndex] = useState(null);
 
   let explorerPrefix: string | undefined;
-  if (web3Provider?.chainId === 1) {
+  if (rpcProvider?._network?.chainId === 1) {
```
shahthepro, line 29: May be, update the hook that returns the provider to also return the chainId? 

This looks a bit hackish because variables and functions prefixed with _ are usually treated as private scoped

shahthepro (commented):
Awesome work, a few comments inline

client/utils/useFutureBlockTime.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  useEffect(() => {
+    const fetchFutureBlock = async () => {
+      let api = `https://api.etherscan.io/api?module=block&action=getblockcountdown&blockno=${blockNumber}&apikey=${ETHERSCAN_API_KEY}`;
+      if (4 === envNetwork) {
+        api = `https://api-rinkeby.etherscan.io/api?module=block&action=getblockcountdown&blockno=${blockNumber}&apikey=${ETHERSCAN_API_KEY}`;
+      }
+      const res = await fetch(api);
+      const futureBlock = await res.json();
+      if (futureBlock?.status !== "0") {
```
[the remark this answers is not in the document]
tomhirst: The default state of `futureBlockTime` is 0, so if the call fails, it'll never update. Then in the UI, if this variable is 0, it'll not show the row in the table. Good call on catching the errors - have wrapped in a try/catch.

client/utils/index.tsx
```diff
@@ -150,3 +151,44 @@ export function claimIsOpen() {
...
+ * Break up proposal content for display
+ *
+ * Important: The output of the function must be
+ * properly escaped and safe to use as raw HTML
+ * that is rendered by UI components. Never return
+ * unsanitized HTML here.
+ */
+const getCleanProposalContent = (proposalDescription: string) => {
+  const split = proposalDescription?.split("<br>");
+  const title = split && split[0];
```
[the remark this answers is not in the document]
tomhirst: Just tested this and `description` returns an empty string if no `<br>` is found in the master string. I think this is OK as we won't render empty vars.

client/utils/index.tsx
```diff
@@ -150,3 +151,44 @@ export function claimIsOpen() {
...
+export const toDaysMinutesSeconds = (totalSeconds: Number) => {
+  const minutes = Math.floor((totalSeconds % 3600) / 60);
+  const hours = Math.floor((totalSeconds % (3600 * 24)) / 3600);
+  const days = Math.floor(totalSeconds / (3600 * 24));
+
+  const minutesStr = makeHumanReadable(minutes, "min");
+  const hoursStr = makeHumanReadable(hours, "hr");
+  const daysStr = makeHumanReadable(days, "day");
+
+  return `${daysStr}${hoursStr}${minutesStr}`.replace(/,\s*$/, "");
```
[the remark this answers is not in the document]
tomhirst: `trim()` would not go far enough because we need to take away the trailing `,` too.

client/pages/proposals/index.tsx
```diff
@@ -58,12 +61,20 @@ const Proposal: NextPage = ({ proposalCount, proposals }) => {
           displayId: proposals.find(
             (p) => p.proposalId.toString() === d.id.toString()
           )?.id,
+          description: proposals.find(
```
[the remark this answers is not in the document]
tomhirst: Good catch, this is messy. Have fixed.

client/components/proposal/ProposalDetail.tsx
tomhirst: Have added a render check to make sure `lastTx` is present.

client/pages/proposals/index.tsx
```diff
@@ -18,13 +17,17 @@ export async function getServerSideProps({ res }: { res: any }) {
...
+    await prisma.proposal.findMany({
+      orderBy: { id: "desc" },
+      include: { transactions: true },
+    })
   ).map((p) => ({
     id: p.id,
     proposalId: p.proposalId,
     createdAt: p.createdAt.toString(),
+    description: p.description,
+    transactions: JSON.parse(JSON.stringify(p.transactions)),
```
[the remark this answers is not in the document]
tomhirst: The `createdAt` field on transaction objects can't be serialized as JSON without this, which is a requirement of returning the data in `getServerSideProps`. Have added a comment!

Screenshot 2022-08-30 at 17 26 54

client/listener.ts
```diff
@@ -209,21 +293,83 @@ const handleEvents = async (blockNumber, events, done) => {
...
+                status: proposalStatus,
+                transactions: {
+                  create: [{
+                    event: statusLabels[proposalStatus],
+                    createdAt: new Date(block.timestamp * 1000),
+                  }],
+                },
+              },
+            });
+            logger.info(`Added ${statusLabels[proposalStatus]} timestamp to proposal ${proposal.proposalId}`);
```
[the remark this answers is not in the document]
tomhirst: I kind of prefer `info` here to `debug` if following these conventions:

Screenshot 2022-08-30 at 17 32 58

Ref: http://www.beefycode.com/post/Log4Net-Tutorial-pt-1-Getting-Started.aspx

client/listener.ts
```diff
@@ -209,21 +293,83 @@ const handleEvents = async (blockNumber, events, done) => {
...
+          } catch(e) {
+            logger.info(e);
+          };
+        }
+      });
+    } else {
+      logger.info('No proposals to update.');
+    }
+  } catch(e) {
+    logger.info(e);
```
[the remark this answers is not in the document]
tomhirst: Agree. Have changed all error messages to `logger.error(e)`.

client/components/proposal/ProposalTable.tsx
tomhirst: It's actually still an object because it has a `0` key so it's calling in alternative notation. I can see how this is confusing to read though, so I've made it clearer that it's an object using bracket notation.

client/components/proposal/SupportTable.tsx
tomhirst: Updated component to catch!

client/components/proposal/ProposalVoteStats.tsx
tomhirst: Good catch!

client/components/proposal/ProposalTable.tsx
tomhirst: Good spot, there shouldn't be any reason. The optional chaining notations were in the wrong place.

client/components/proposal/ProposalActionsTable.tsx
tomhirst: Re-written to make it look cleaner.

tomhirst:
I gave the "register to vote" CTA some thought overnight and devised this solution. I've re-used an existing CTA pattern to display the section to users with veOGV-positive wallets that haven't self-delegated. This will appear on the home page (last 5 proposals) and the proposals index page. 

LMK if you think anything further is needed @matt-bullock and if you'd like to revise the copy at all @micahalcorn.

Screenshot 2022-09-02 at 11 00 37

Screenshot 2022-09-02 at 11 00 44

tomhirst:
I've included the CTA on the single proposal view too, thanks for the recommendation @matt-bullock 

Screenshot 2022-09-02 at 11 46 57

Issue #325, opened 2022-08-30
Wrong time remaining on stake


micahalcorn:
Today is **Aug 30, 2022** and this seems to indicate that my lock-up period will end in 8 days. In fact, I believe that it ended 8 days _ago_. In this case, we could replace that text with "None" for the second column. I think I would also replace the heading on the third column with "End of lock-up".

<img width="574" alt="Screen Shot 2022-08-30 at 8 54 40 AM" src="https://user-images.githubusercontent.com/273937/187483920-23abd328-f863-4d47-a233-c9002a214f7f.png">

PR #327, opened 2022-08-31
Fix time diff and rename

0xcodercrane:
This resolves the issue https://github.com/OriginProtocol/ousd-governance/issues/325

### DIFF
--- a/client/components/vote-escrow/LockupsTable.tsx
+++ b/client/components/vote-escrow/LockupsTable.tsx
@@ -66,7 +66,7 @@ const LockupsTable: FunctionComponent = () => {
           <tr>
             <th>OGV</th>
             <th>Time remaining</th>
-            <th>End of lock-up</th>
+            <th>Lockup ends</th>
             <th>veOGV</th>
           </tr>
         </thead>
--- a/client/components/vote-escrow/LockupsTable.tsx
+++ b/client/components/vote-escrow/LockupsTable.tsx
@@ -66,7 +66,7 @@ const LockupsTable: FunctionComponent = () => {
           <tr>
             <th>OGV</th>
             <th>Time remaining</th>
-            <th>Lockup ends</th>
+            <th>End of lock-up</th>
             <th>veOGV</th>
           </tr>
         </thead>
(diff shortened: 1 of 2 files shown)


shahthepro (approved):

client/components/vote-escrow/LockupsTable.tsx
```diff
@@ -66,7 +66,7 @@ const LockupsTable: FunctionComponent = () => {
           <tr>
             <th>OGV</th>
             <th>Time remaining</th>
-            <th>Lockup ends</th>
+            <th>End of lock-up</th>
```
tomhirst, line 69: We would also need the conditional statement here to ensure we render the right heading for the row.

tomhirst (changes requested):
Thanks for taking a look. Just a small comment @0xcodercrane.

client/components/vote-escrow/LockupsTable.tsx
0xcodercrane: I think if you can make 100% sure the table has only 1 row, conditional statement should be implemented. If the lockups table might have more than 2 rows, this will not make sense. 
But I guess the lockup contract looks similar to curve voting escrow which allows only 1 lock per address. If this is our case, I will add conditional statement as well.

tomhirst: Ah, yes, you're right, our contracts allow multiple lock ups so we need the heading to make sense no matter the value in the row(s) beneath.

Suggestion: 

Instead of "None" we could use "-" which, alongside the Unstake button becoming active, will make it obvious which stakes have no duration remaining at a glance.

0xcodercrane: Nice! do I have to keep `End of lock-up` or revert it `Lockup ends` ?

tomhirst: I would revert 🤝

0xcodercrane: Resolved in 944cfdebe344e4d2a04f4f62f6bb550764f562b0

tomhirst (approved):
LGTM - merging 🚀

Issue #324, opened 2022-08-27
Implement governance parameters for launch


micahalcorn:
See https://github.com/OriginProtocol/ousd-governance/issues/295#issuecomment-1229254258 ☑️

Issue #295, opened 2022-08-02
Agree on governance parameters


micahalcorn:
Let's take inventory of each parameter that controls the on-chain governance system. This includes voting period, timelock, quorum requirement, proposal threshold, etc.

joshfraser:
It would be helpful to take inventory of what numbers other established governance systems are using. For example, I know Compound requires 0.25% to issue a vote and 4% for a quorum. What about other protocols? Let's learn from those who have gone before. My initial reaction is that 4% feels rather low from a security standpoint.

Techmoen:
In simple terms, at 4%, around 0.4% of the veOGV balance in OGV would be required to reach quorum currently. At the current level of 9.7B veOGV, this would require (1/10.5)*4%*9.7B =  ~37M OGV. With 478M OGV in circulation currently, this is approximately 8% of the circulating supply, the main question is how much of this circulating supply is liquid (ie. on exchanges).

If there were to be a coordinated governance attack, purchases of OGV would occur on a DEX to avoid KYC. Based on Uniswap liquidity alone, 4% would be secure as not enough liquidity currently exists there. However, as someone buys in on Uniswap, market makers will step in and help satisfy the liquidity required on the DEX. If someone were to slowly accumulate OGV in this way, then you have to assume total liquidity across all exchanges (CEX and DEX) would be in play. From this, you can then estimate the slippage and arrive at a total cost to reach quorum.

For example: assuming about half of the circulating supply is liquid (~240M OGV), and that to accumulate 37M OGV would incur total price slippage of 500%, it would cost approximately $0.007145*5*37M = $1.32M to reach quorum. This would be fairly expensive with no guarantee of success, as the Origin team will have other controls in place.

micahalcorn:
@Techmoen thanks for digging into this. Circulating supply is actually almost double the figure that you have here. I suspect that you referenced CoinGecko or CoinMarketCap, which were wrong. We're working on getting those updated. The source of truth will be our own endpoint at https://www.originprotocol.com/circulating-ogv.

micahalcorn:
**Current Compound parameters**

|name|value|
|----|---|
|Voting delay|13,140 blocks (2 days)|
|Voting period |19,710 blocks - (3 days)|
|Proposal threshold |25,000 COMP ($1.5M, 0.37% of circulating supply)|
|Quorum votes|400,000 COMP ($23.5M, 5.83% of circulating supply)|
|Timelock|2 days|

[...]

PR #328, opened 2022-08-31
Update governance parameters

tomhirst:
Addresses: #324

### DIFF
--- a/contracts/Governance.sol
+++ b/contracts/Governance.sol
@@ -19,13 +19,13 @@ contract Governance is
         Governor("OUSD Governance")
         GovernorSettings(
             1, /* 1 block */
-            32727, /* 5 days */
-            5000000 * 1e18 /* 5 mio veOgv */
+            17280, /* ~3 days (86400 / 15) * 3 */
+            10000000 * 1e18 /* 10 mio veOgv */
         )
         GovernorVotes(_token)
-        GovernorVotesQuorumFraction(4) // Default quorum denominator is 100, so 4/100 or 4%
+        GovernorVotesQuorumFraction(20) // Default quorum denominator is 100, so 20/100 or 20%
         GovernorTimelockControl(_timelock)
-        GovernorPreventLateQuorum(86400 / 15) // ~1 day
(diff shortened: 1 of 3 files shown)


contracts/Governance.sol
```diff
@@ -19,13 +19,13 @@ contract Governance is
         Governor("OUSD Governance")
         GovernorSettings(
             1, /* 1 block */
-            32727, /* 5 days */
-            5000000 * 1e18 /* 5 mio veOgv */
+            17280, /* ~3 days (86400 / 15) * 3 */
```
franckc, line 22: These days the average block time is more in the order of 13 sec. See https://ycharts.com/indicators/ethereum_average_block_time
So that would give a slightly larger number of blocks: 86400 / 13 * 3 = 19938

The initial code had used a value of 13.2sec to get to 32727 blocks = 5 days.

Though the calculation for GovernorPreventLateQuorum below was using 15sec in the old code. So I guess it was inconsistent...

Anyway it's an approximation. I think what you have of using 15sec and consistently using it for both settings is good.

franckc (approved):
Verified the new settings match what we agreed on in https://github.com/OriginProtocol/ousd-governance/issues/295#issuecomment-1229254258

LGTM

shahthepro (approved):

contracts/Governance.sol
tomhirst: > These days the average block time is more in the order of 13 sec. See https://ycharts.com/indicators/ethereum_average_block_time So that would give a slightly larger number of blocks: 86400 / 13 * 3 = 19938
> 
> The initial code had used a value of 13.2sec to get to 32727 blocks = 5 days.
> 
> Though the calculation for GovernorPreventLateQuorum below was using 15sec in the old code. So I guess it was inconsistent...
> 
> Anyway it's an approximation. I think what you have of using 15sec and consistently using it for both settings is good.

Exactly my thinking. I was tempted to go with the current block mine time but noticed 15 seconds being used previously so I standardised everything.

PR #331, opened 2022-09-09
Add burn page

tomhirst:
Screenshot 2022-09-09 at 19 57 31

### DIFF
--- a/client/components/TokenAmount.tsx
+++ b/client/components/TokenAmount.tsx
@@ -6,9 +6,14 @@ import { useStore } from "utils/store";
 interface TokenAmount {
   amount: BigNumber | string | number;
   format?: string;
+  isWalletBalance?: Boolean;
 }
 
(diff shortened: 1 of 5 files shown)


micahalcorn:
44.03 + 78.90 != 66.77

The 66.77% is referring to 765M / 1.45B but it looks like the sum of claims. Maybe just remove the percentages from 175M and 590M? cc @matt-bullock

micahalcorn:
> 44.03 + 78.90 != 66.77
> 
> The 66.77% is referring to 765M / 1.45B but it looks like the sum of claims. Maybe just remove the percentages from 175M and 590M? cc @matt-bullock

Maybe it's okay to keep them. The 66.77 is basically a weighted average instead of a sum. People can figure it out.

shahthepro (approved):
Took a quick pass, looks good to me

PR #336, opened 2022-09-17
add exchange airdrop allocation to burn calculation

rolandpo:
Including portion of tokens sent to exchanges in claim calculation

### DIFF
--- a/client/pages/burn.tsx
+++ b/client/pages/burn.tsx
@@ -75,30 +75,22 @@ const Burn: NextPage = () => {
     totalAirdropped
   );
 
-  const mandatoryDistributorInitialOgv =
-    BigNumber.from(398752449).mul(decimal18Bn);
-  const optionalDistributorInitialOgv =
-    BigNumber.from(747905084).mul(decimal18Bn);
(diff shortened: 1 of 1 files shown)


tomhirst:
> Including portion of tokens sent to exchanges in claim calculation

I'm not sure about this one.

I like using the initial supplies of the contracts in the calculations as we can assure any remaining there will be burned. As far as I was last aware, exchanges might honour the burn, but they might not.

I've reviewed the code and it looks good, but I'll let the product team take the lead here. cc: @micahalcorn

rolandpo:
Yep no worries, just an idea. Was proposing this since there was some confusion in discord, where someone assumed 67% claimed meant 33% would be burned.

Just to reiterate, this change would assume that exchanges don't burn anything, so that there's no unpleasant surprises if that actually happens. Also it would help fix the discrepancy between the Airdrop tab on the left and the Claim page on the right

micahalcorn:
Per https://github.com/OriginProtocol/origin-dollar/issues/1112, I'm going to close this and also recommend that we redirect this burn page to the new one.

PR #343, opened 2022-09-23
Catch erroneous calldata error and return empty array to UI

tomhirst:


sparrowDom (approved):
LGTM

DanielVF:
This could be dangerous if it shows a proposal doesn't have actions when it does. Just because these signatures are invalid  does not mean the 4byte hash of the signature is invalid at the solidity level.

PR #350, opened 2022-10-10
remove the check for claims being open

sparrowDom:
On a couple of places we had a check that didn't start interacting with contracts until the staking Claims were not open. This PR removes that check


sparrowDom:
because of this being broken on production the changes in this PR are already deployed. Happy to address anything if necessary later

tomhirst (approved):
Approving this because the check is no longer relevant now we're in a post-claim world.

PR #359, opened 2022-11-10
add posibility to self delegate to register for voting

sparrowDom:
add possibility to self delegate to register for voting. If user hasn't self delegated yet this link is shown: 

Screenshot 2022-11-10 at 00 53 08

takes user to this page: 
Screenshot 2022-11-10 at 01 49 34

in case user hasn't locked up veOgv yet this is shown: 
Screenshot 2022-11-10 at 01 50 46


sparrowDom:
Merged this in and deploying it (since it is relatively straight forward change). Happy to do any adjustments subsequently.

PR #361, opened 2022-11-21
Increase extend gas allowance

DanielVF:
This moves gas sent for `extend` transactions up from a hardcoded 195,000 up to 220,000.

This should fix the failing transactions on extends.


sparrowDom (approved):
LGTM

Issue #312, opened 2022-08-05
Remove voter registration requirement


micahalcorn:
We should fix this so that users are not forced to submit a transaction prior to being able to vote or create a proposal.

<img width="1044" alt="Screen Shot 2022-08-05 at 8 16 40 AM" src="https://user-images.githubusercontent.com/273937/183108000-dbdc71a6-72c3-4d60-8685-8c78504f97d4.png">

DanielVF:
Doing this at the contract level will:

A: Bump the cost of staking up by 25K gas for first stake and 3K or so for subsequent stakes
B: We'll still need some kind of UI, since not everyone will have it.

We'll want a delegation UI regardless in the long run. The % of stakes who vote is a pretty small number.

micahalcorn:
Just updating this to say that we've decided to hold off on making the contract changes until we have more of a reason to encourage voter participation 🗳️

micahalcorn:
Updating this again to say that we're going to make governance participation a higher priority in the new year and will want to clean this up so that new stakers have a smoother experience 🪒

franckc:
@sparrowDom Could you help with deploying the associated contract changes?

sparrowDom:
thanks for flagging this @franckc [this is the proposal](https://governance.ousd.com/proposals/106247227237600433139581351842615975456822599329363014399616832913328325135207) and it has [just been executed](https://etherscan.io/tx/0x555f1b564ac5c86b454736b630d44855db3538912a63ce77201815296e4cba40)

PR #367, opened 2023-01-30
Auto-delegate vote power on stake

shahthepro:
Related issue: #312 

- [x] Delegates to self after staking
- [x] Changes the CI workflow a bit and also runs Forge tests on CI now
- [x] Adds deployment script to upgrade `OGVStaking` contract

I left out removing the client component we have right now intentionally, because:
- It has a banner that says something like "Stake OGV to vote" 
- The delegate CTA could still be useful if someone staked on the older contract or managed to call `delegate(address(0))` directly on contract for whatever reasons

Questions:
- Should `extend` also check and delegate to self?

-----

Contract change checklist:
  - [x] Unit tests pass


.github/workflows/main.yaml
```diff
@@ -27,14 +32,45 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Cache Compiler Installations
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.solcx
```
sparrowDom, line 59: curious does this preserve the build artefacts from previous compilations?

shahthepro: Actually no, this is for solidity compiler. But we cache built artifacts on the origin-dollar repo since we use that a lot for fork tests. Can do the same optimisations here too

sparrowDom (approved):
Looks great

DanielVF:
@micahalcorn Note that adding this will make a noticeable bump up in the gas cost to stake, if it ends up delegating for the user. We should probably convert the dapp to estimate + buffer rather than the current hardcoded values.

shahthepro:
@sparrowDom @DanielVF I changed the extend method to delegate too (if user has never set it before). A bit of testing on fork looks good to me, marking this as ready for review. Let me know if you have any comments

sparrowDom:
LGTM. I would just add a comment why the manualDelegation mapping has been added. Just that we don't forget about it in the future.

shahthepro:
Merging this with master, since I there are a few fixes and changes to CI as well. Will address any comments in other PRs

contracts/OgvStaking.sol
```diff
@@ -151,6 +156,12 @@ contract OgvStaking is ERC20Votes {
         );
         _mint(to, points);
         ogv.transferFrom(msg.sender, address(this), amount); // Important that it's sender
+
+        if (!hasManualDelegation[msg.sender] && delegates(to) == address(0)) {
```
DanielVF, line 160: 🛑 Using `message.sender` in this line is a bug?

The scenario we are trying to avoid is:

- An address renounces delegation.
- A third party gives that address a tiny stake, and reenables delegation. 

If we do the `hasManualDelegation`check on the sender of the stake, not the receiver, even if the receiver has revoked voting, as long as the attacker has not delegated, the attacker can still complete the attack.

This test fails in forge.

```javascript
    function testRenounceAttack() external {
        // Alice can vote, because she is staked
        assertEq(staking.getVotes(alice), 1 * POINTS, "can vote after staking");

        // Alice renounces voting.
        vm.prank(alice);
        staking.delegate(address(0));

        // Attacker attacks
        vm.startPrank(attacker);
        ogv.mint(attacker, 1 ether);
        ogv.approve(address(staking), 1 ether);
        staking.stake(1 ether, 100 days, alice);
        vm.stopPrank();

        vm.roll(2);

        // Alice should still have renounced voting
        assertEq(staking.getVotes(alice), 0, "can't vot after renouncing");
    }
 ```

File: contracts/OgvStaking.sol [hunk 3]
```diff
@@ -33,6 +33,11 @@ contract OgvStaking is ERC20Votes {
     mapping(address => uint256) public rewardDebtPerShare;
     uint256 public accRewardPerShare; // As of the start of the block
 
+    // Used to track any manual `delegate()` method call. When this isn't
+    // set to true, voting powers are delegated to the receiver of the stake
+    // when `stake()` or `extend()` method are called.
+    mapping(address => bool) public hasManualDelegation;
```
DanielVF, line 39: "Manual" may be the wrong word in this variable name, since almost everyone with this set will have had it automatically set by via this new code?

contracts/OgvStaking.sol
```diff
@@ -280,4 +295,14 @@ contract OgvStaking is ERC20Votes {
...
         emit Reward(user, netRewards);
     }
+
+    /**
+     * @dev Change delegation for `delegator` to `delegatee`.
+     *
+     * Emits events {DelegateChanged} and {DelegateVotesChanged}.
+     */
+    function _delegate(address delegator, address delegatee) internal override {
+        hasManualDelegation[delegator] = true;
```
sparrowDom, line 305: On second pass, this should probably get set only when user explicitly calls `delegate`. And rather than being set here I think you will need to override [this function](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Votes.sol#L143-L145) and set it there

contracts/OgvStaking.sol
sparrowDom: I have overlooked this 😞 I think you are right @DanielVF it should be `!hasManualDelegation[to]` in there.

shahthepro: Thanks for catching this, will fix it now

contracts/OgvStaking.sol
shahthepro: I thought about that but in some cases it may cause weird behaviour if we do that, for instance:
1. Alice stakes and gets self-delegated (so the flag isn't turned to true)
2. Alice stakes again but to Bob (a different address). 
In the above case, since the flag isn't true, the contract might transfer all voting powers of Alice as well to Bob, leaving Alice with no voting power. 

I can probably think of a solution to make it work but it might make the contract more complicated

contracts/OgvStaking.sol
shahthepro: PR #377 should fix this

contracts/OgvStaking.sol
shahthepro: Resolved in PR #377

contracts/OgvStaking.sol
DanielVF: I think that the updated code is good.

The real problem here isn't about about detecting manual changes, but about knowing if delegation has been set before or not, and this setup handles that, I think.

DanielVF (commented):
This a review for both this PR and PR #377

Action items:
- [ ] Add test for extend auto-stake path
- [x] Confirm that the DAPP gas code is updated
- [x] Add comment on `hasDelegationSet` that because of upgrades, some users may have a delegation but not have hasDelegationSet true.

#### Requirements

We want to have users self delegated for voting the first time they receive staked funds.

We don't want users to be able to have their delegated shares changed by others gifting stakes to them.

It looks like this *will add a 92,500 gas cost* to a user's first stake, which is almost a 50% cost increase to staking. 😢 

### Deployment Considerations

- [x] ❗ The dapp must support these higher gas prices, or all staking from the UI will be broken.

#### Internal State

- ✅ After any stake, if the user has not delegated votes before, they will be delegated and able to vote.
- ✅ hasDelegationSet can only move from false to true
- hasDelegationSet will be false on old users with delegation set
- 🟦 hasDelegationSet will remain false after a stake on old users that already have delegation set. (This may need a comment)

#### Attack

The biggest concern is that someone alter someone else's stake, with the worse situation being able to delegate another user's stake to the attacker.

✅ A key condition then is that the "from" and the "to" of the delegate() call always be the same user. This is currently true.

#### Logic

- [x] Logic good

#### Tests

  - [x] Each publicly callable method has a test
  - [ ] Each logical branch has a test

❗ We never test the auto-delegate on `extend`. This is partly because in our test environment all users have a delegation as soon as they are staked, and staking must be done before extending. This doesn't match our production setup, with staked users that do not have delegation set.

It's possible to use some fancy forge features to undo the auto-delegate for a test, so we can test this.

```javascript
    function testDelegateOnExtend() external {
        // For test purposes, undo auto-staking on user
        vm.prank(oak);
        staking.delegate(address(0));
        stdstore.target(address(staking))
            .sig(staking.hasDelegationSet.selector)
            .with_key(oak)
            .checked_write(false);

        vm.roll(1);
        
        // Cannot vote because test undid auto-staking
        assertEq(staking.getVotes(oak), 0, "can not vote");
        assertEq(staking.delegates(oak), address(0), "no delegation");
        assertEq(staking.hasDelegationSet(oak), false, "no hasDelegationSet");

        // Extend should auto-delegate
        vm.prank(oak);
        staking.extend(0, 200 days);
        assertEq(staking.delegates(oak), oak, "should auto delegate on extend");
        assertGt(staking.getVotes(oak), 1 * POINTS, "should have voting power after extend");
    }
```

#### Flavor

_Flavor good._

#### Overflow

_No overflow options._

#### Proxy
- [x] Make sure proxy implementation contracts don't initialize variable state on variable declaration and do it rather in initialize function.

#### Black magic

_No magic_

#### Dependencies
- [x] Reviewed `_delegate` dependency in OZ ERC20Votes.sol

#### Deploy
- [x] Check that any deployer permissions are removed after deploy

#### Authentication

- [x] Never use tx.origin
- [x] Check that every external/public function should actually be external
- [x] Check that every external/public function has the correct authentication

#### Cryptographic code

_No Crypto code_

#### Gas problems

_No gas problems_

#### External calls

_No external calls in change_

#### Ethereum

PR #370, opened 2023-01-31
Sparrow dom/UI work

sparrowDom:
- separate create new proposal button into a component, that shows "not sufficient voting power", when user has too little veOGV to create a proposal
- removed Reallocation UI part of new proposal creation since it is out of date


shahthepro (approved):
LGTM

PR #371, opened 2023-02-02
Fixed veOgv bug on dApp

### LINKED ISSUE #342: Unexpected veOGV amount
I just went through the process of creating a new stake. I'm 99% sure that I was told I would receive ~830k veOGV, but [I ended up getting ~930k](https://etherscan.io/tx/0xd3360139c1a05653cd1731eb444f8bb632a1b336a2ade5c004c2e1c77e1e4525). I'll watch for this again next time I go through the experience. Let me know if someone else is able to reproduce it first.

HrikB:
My diagnosis of this bug was two fold.

The first is that the code for calculating the veOgv in the smart contract is quite different from the one used in the dApp.

`OgvStaking.sol`
```
function previewPoints(uint256 amount, uint256 duration)
        public
        view
        returns (uint256, uint256)
    {
        // Basically always block.timestamp
        uint256 start = block.timestamp > epoch ? block.timestamp : epoch;
        uint256 end = start + duration;
        uint256 endYearpoc = ((end - epoch) * 1e18) / 365 days;
        uint256 multiplier = PRBMathUD60x18.pow(YEAR_BASE, endYearpoc);
        return ((amount * multiplier) / 1e18, end);
    }
```

`LockupForm.tsx`
```
const veOgvFromOgvLockup =
    lockupAmount * votingDecayFactor ** (lockupDuration / 12);
```

The typescript version yields quite a similar output but it has a lot of rounding errors that often would go unnoticed because it's not huge; but often wouldn't. So changed that to follow the smart contract's calculation method.

The second and bigger issue (and the one that was the crux of the bug @micahalcorn found) came from a use of a static `block.timestamp`. The `blockTimestamp` used in the dApp is one that is fetched in the initial page load and then never updated. This is a problem because the `block.timestamp` is part of the calculation for the output veOgv quantity. 

If a user were to go on the page, leave for an hour and come back to confirm their transaction... they would find a very different veOgv output quantity.

The change made here is to wrap an already existing block timestamp fetching hook in a `useInterval()` delimited at 4s. At an average 15s block production time, this is more than enough to ensure a non-stale timestamp.

One thing to note is the formula in dApp will use the *current* block timestamp as opposed to the next (the one in which the user's transaction will be confirmed). However, the difference in output is negligible. 

Closes #342


client/utils/useBlock.tsx
```diff
@@ -16,13 +17,20 @@ const useBlock = () => {
...
 
+    let intervalId: ReturnType<typeof setInterval>;
+
     if (web3Provider && networkInfo.correct) {
-      Promise.all([getBlockTimestamp()]).then(([blockTimestamp]) => {
-        useStore.setState({
-          blockTimestamp,
+      intervalId = setInterval(() => {
+        Promise.all([getBlockTimestamp()]).then(([blockTimestamp]) => {
+          console.log(blockTimestamp);
```
shahthepro, line 25: nit: can remove these logs before merging

client/components/vote-escrow/LockupForm.tsx
```diff
@@ -145,8 +145,15 @@ const LockupForm: FunctionComponent<LockupFormProps> = ({ existingLockup }) => {
   // as specified here: https://github.com/OriginProtocol/ousd-governance/blob/master/contracts/OgvStaking.sol#L21
   const votingDecayFactor = 1.8;
 
-  const veOgvFromOgvLockup =
-    lockupAmount * votingDecayFactor ** (lockupDuration / 12);
+  const epoch = 1657584000;
```
shahthepro, line 148: nit: A comment on where this timestamp came from would be helpful

shahthepro (approved):

PR #373, opened 2023-02-03
Remove deploy from master

shahthepro:
The staging app is configured on heroku to deploy from `staging` branch. So, deploying `master` to staging as well seems redundant and since it's a simple push, we don't have to have it in a workflow file


sparrowDom (approved):
LGTM

PR #377, opened 2023-02-07
Check delegation flag on receiver

shahthepro:

Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/contracts/OgvStaking.sol
+++ b/contracts/OgvStaking.sol
@@ -35,8 +35,9 @@ contract OgvStaking is ERC20Votes {
 
     // Used to track any calls to `delegate()` method. When this isn't
     // set to true, voting powers are delegated to the receiver of the stake
-    // when `stake()` or `extend()` method are called. For existing stakers 
-    // with deletegation set, This will always remain `false` (due to upgrades)
+    // when `stake()` or `extend()` method are called. 
+    // For existing stakers with deletegation set, This will remain `false` 
+    // unless the user calls `delegate()` method.
     mapping(address => bool) public hasDelegationSet;
 
     // Events
--- a/contracts/OgvStaking.sol
+++ b/contracts/OgvStaking.sol
@@ -33,10 +33,10 @@ contract OgvStaking is ERC20Votes {
     mapping(address => uint256) public rewardDebtPerShare;
     uint256 public accRewardPerShare; // As of the start of the block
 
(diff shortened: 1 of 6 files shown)


DanielVF:
Here's the final PR review on this PR and the previous one. Three action items. Again, thanks for all the tests!

https://github.com/OriginProtocol/ousd-governance/pull/367#pullrequestreview-1289429916

sparrowDom (approved):
LGTM thanks for the extra test @shahthepro. 

Regarding the dapp gas increases would you mind spinning up these contracts on the forked node and see how much `extend` & `stake` are gas intensive? We have left some ~20%-30% hardcoded dapp gas margin on those calls, and if there is still ~15% gas margin after these changes we might not need to increase the gas limits.

contracts/OgvStaking.sol
```diff
@@ -33,10 +33,11 @@ contract OgvStaking is ERC20Votes {
     mapping(address => uint256) public rewardDebtPerShare;
     uint256 public accRewardPerShare; // As of the start of the block
 
-    // Used to track any manual `delegate()` method call. When this isn't
+    // Used to track any calls to `delegate()` method. When this isn't
     // set to true, voting powers are delegated to the receiver of the stake
-    // when `stake()` or `extend()` method are called.
-    mapping(address => bool) public hasManualDelegation;
+    // when `stake()` or `extend()` method are called. For existing stakers 
+    // with deletegation set, This will always remain `false` (due to upgrades)
```
sparrowDom, line 39: This comment should be updated, since even if existing users call `extend`, `stake`, or `delegate` this value for them should change to true. As demonstrated in this test: https://github.com/OriginProtocol/ousd-governance/pull/377/commits/7287c88635bd6dd1a0d8e26d9378a88d4931b33a

Other than that LGTM

sparrowDom (approved):

contracts/OgvStaking.sol
shahthepro: Actually, calling `extend` or `stake` won't change the boolean when the user has delegation set already. But `delegate` will switch the flag. Will update the comment

shahthepro:
@DanielVF @sparrowDom The action items and comments from the previous PR are all addresses now, including using a gas estimate. We could use the `hasDelegationSet` method from the contract to determine a different hardcoded lower limit but I'm leaving that to another PR (after deployment) since it'll fail on mainnet right now

PR #378, opened 2023-02-21
OGV staking deploy

sparrowDom:
OGV staking deploy

Implementation contract [transaction](https://etherscan.io/tx/0xaff675180f42be8356fa603a1d059bbbf0b962b4323351a2ad66cd0193f060ea) and [address](https://etherscan.io/address/0xe61110663334794aba03c349c621a075dc590a42)

|Name|Address|
|--|--|
|Staking Implementation| 0xe61110663334794aba03c349c621a075dc590a42|

This PR also adds tooling so that OGV Governor proposal can more easily be built.


shahthepro (approved):

common.py
```diff
@@ -0,0 +1,52 @@
+import os
+from brownie import *
+import brownie
+from ape_safe import ApeSafe
```
DanielVF, line 4: We can remove this, right, since we aren't using ape safe?

DanielVF (approved):

DanielVF:
I note that this will also set the name of the contract to "veOGV".

- [x] Deployed contract verified code (and dependancies) match the final PR commit's code
- [x]  The transactions that interacted with the newly deployed contract match the deploy script.
- [x]  Governance proposal matches the deploy script
- []  Smoke tests pass after fork test execution of the governance proposal

common.py
sparrowDom: yup thanks, removed

PR #379, opened 2023-03-03
Using @web3react/core for wallet sign-in

HrikB:
The main change in this PR is the overhaul of the wallet system but I also made some other tangential changes to accommodate	 a direct Ledger login and removal of duplicate data. 

First, I made some changes to the global store. The way that governance currently works with the contracts in the store is that it defaults to attaching a read-only provider and when a wallet is connected, the provider in every contract is upgraded to a signer. However, this for some reason poses an issue to the Ledger connector which throws an error saying it won’t allow multiple connections. So now, there is no upgrade to the contract instances in the store when a wallet logs in. Instead, whenever a state-changing call is made to Ethereum, the signer is manually connected to the contract instance which seems to make the Ledger connector happy.

There also seems to be three different providers being put in the store… I don’t see any reason for this so I consolidated it all into `web3Provider`. `rpcProvider` and `provider` were removed from the store. In addition to that, I removed the address and chainId keys from the store. They are unnecessary since `chainId` and `account` can be retrieved from `useWeb3React()`.

I also removed some components from the repo. I got rid of the Eligibility component since it doesn’t seem to be used anywhere as well as the Web3Button component since it has been replaced with the new ConnectButton component which much more closely follows the signing in pattern from origin-dollar dapp. Also seemed like a lot of the components for the claim route were unused… so I removed them.

Tasks Left:
- [ ] Eager Connect


smitch88:
Deprecated

PR #390, opened 2023-07-04
Use `previewPoints` on contract to compute APY

### LINKED ISSUE #342: Unexpected veOGV amount

shahthepro:
- Gets the veOGVReceived from contract (Fixes #342)
- Fixes under-reporting of APY calculation (Need someone to verify the APY)


client/utils/useStakingAPY.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const [loading, setLoading] = useState(true);
+  const [veOgvReceived, setVeOGVReceived] = useState(0);
+
+  const { totalSupplyVeOgvAdjusted } = totalBalances;
+
+  useEffect(() => {
+    if (!contracts.loaded) return;
+
+    let done = false;
```
smitch88, line 18: Not familiar with this pattern for how you are using `done`, what are you trying to solve? Just running the `go` function once?

client/pages/claim.tsx
```diff
@@ -41,9 +34,9 @@ const ClaimPage: NextPage<ClaimPageProps> = () => {
             The claim period has now expired
           </h2>
           <div className="mt-3 text-sm font-bold">
-            {`All unclaimed OGV was burned after 90 days. You can still buy OGV and stake it for up to four years. Current OGV stakers are earning a vAPY of ${stakingApy.toFixed(
-              2
-            )}%.`}
+            {`All unclaimed OGV was burned after 90 days. You can still buy OGV and stake it for up to four years. Current OGV stakers are earning a vAPY of ${
+              apyLoading ? "--.--" : stakingAPY.toFixed(2)
```
sparrowDom, line 38: nit: missing a % here. Should be:
```
apyLoading ? "--.--" : `${stakingAPY.toFixed(2)}%`
```

sparrowDom (approved):
great job at simplifying it and making it correct 💪

client/utils/useStakingAPY.tsx
shahthepro: Just defensive programming. `go` is an async function. It's possible that the component might rerender before the function completes execution. If it does, then `setState` will throw a bunch of console errors because it's trying to update state of a detached component. When the component demounts, done is set to true and the state isn't updated in that case

client/pages/claim.tsx
shahthepro: Fixed. Thanks

PR #406, opened 2023-11-20
Rebrand OGV contracts and tweak Governance parameters

shahthepro:
Adds deploy_005 script that when executed will:
- Deploy rebranded OGV and veOGV contracts and upgrade their proxies
- Changes `votingPeriod` to 2 days with Blocks per seconds set to 12 (From the current value of 3 days with 15 bps which is way off)
- Changes `lateQuoromVoteExtension` to 1 day (down from current value of 2 days)

Also, updates the `common.py` file to simulate execution of the proposal on fork

------
### New Contracts

| Name | Address |
|----|----|
| OgvStaking | [0xfd8e7dacc16c3807e66fedd4f7c65f26b1fee558](https://etherscan.io/address/0xfd8e7dacc16c3807e66fedd4f7c65f26b1fee558) |
| OriginDollarGovernance | [0x190140d8d1815f6f92fd58df0dacb758276fffdb](https://etherscan.io/address/0x190140d8d1815f6f92fd58df0dacb758276fffdb) |

| - | - |
| ---- | ---- |
| Deployer | [0x58890a9cb27586e83cb51d2d26bbe18a1a647245](https://etherscan.io/address/0x58890a9cb27586e83cb51d2d26bbe18a1a647245) |
| Proposer | [0x6a6d776120f7e4a8dba5f6bf49b85cb340cfe241](https://etherscan.io/address/0x6a6d776120f7e4a8dba5f6bf49b85cb340cfe241) |
| Proposal Tx | [0x9724611ddf5d1fc5e776a8e0c985d64e341857122582ab0fe362c72b445b232e](https://etherscan.io/tx/0x9724611ddf5d1fc5e776a8e0c985d64e341857122582ab0fe362c72b445b232e) |
| Proposal ID | [22983443710848978447682038093447564105657898650006428914638357537618097496145](https://governance.ousd.com/proposals/22983443710848978447682038093447564105657898650006428914638357537618097496145) |

------

Contract change checklist:
  - [ ] Unit tests pass


contracts/GovernanceToken.sol
```diff
@@ -21,7 +21,7 @@ contract OriginDollarGovernance is
     constructor() initializer {}
 
     function initialize() public initializer {
-        __ERC20_init("Origin Dollar Governance", "OGV");
```
DanielVF, line 24: I don't think this is going to change the name, since this is in the initializer, and only called early on.

I think we many just want to add a "name()" function like OGV staking has, and return the correct name there.

shahthepro: That's my bad. I totally missed that it's in the initializer. Will change it

scripts/deploy_005_rebranding.py
```diff
@@ -0,0 +1,72 @@
...
+        },
+        {
+          'contract': ogv_proxy,
+          'signature': 'upgradeTo(address)',
+          'args': [ogv_impl.address]
+        },
+        {
+          'contract': governance,
+          'signature': 'setVotingPeriod(uint256)',
+          'args': [BLOCKS_PER_DAY * 3] # 3 days
```
DanielVF, line 62: Sorry, I've lost track of the discussion around this, but perhaps this should be 2 days? That still gives enough time to vote and to see things, while being able to react to other protocols changing quicker.

shahthepro: Yeah. 2 days makes more sense.

2 days for voting, 1 day late quorum extension, and a 2 day timelock would make it 5 days at max

tests/test_token.py
```diff
@@ -1,60 +1,55 @@
...
 def test_name(token):
-    assert token.name() == "Origin Dollar Governance"
-
+    assert token.name() == "Origin DeFi Governance"
 
 def test_symbol(token):
     assert token.symbol() == "OGV"
 
+def test_staking_name(staking):
+    assert staking.name() == "Vote Escrowed Origin DeFi Governance"
```
sparrowDom, line 12: thanks for adding these tests

common.py
```diff
@@ -45,7 +48,32 @@ def governanceProposal(deployment):
...
+
+            print("Executing proposal")
+            ogvGovernor.execute(proposalId, {'from': GOV_MULTISIG})
+
+
         print("Execute the following transaction to create OGV Governance proposal")
         print("To: {}".format(propose_tx.receiver))
         print("Data: {}".format(propose_tx.input))
 
+def timetravel(seconds):
```
sparrowDom, line 77: nit: maybe just add a comment that this moves the chain up for a certain number of blocks that correspond to voting periods in the contracts, but doesn't actually increase the block timestamps (there is an `evm_increaseTime` RPC for that)

sparrowDom (approved):
LGTM great work

common.py
shahthepro: `brownie.chain.sleep` is actually used in this function to increase the block timestamp as well. [Here's a link](https://eth-brownie.readthedocs.io/en/stable/core-chain.html#time-travel) to the brownie docs

sparrowDom: ah right, thanks 👍

sparrowDom (approved):
The PR and the governance proposal look OK

DanielVF (approved):
Contract changes are good.

PR #408, opened 2024-04-17
Add OGNRewardsSource contract

shahthepro:

Contract change checklist:
  - [ ] Unit tests pass


contracts/OGNRewardsSource.sol
```diff
@@ -0,0 +1,156 @@
...
+
+    /// @dev Collect pending rewards
+    /// @return rewardAmount Amount of reward collected
+    function collectRewards() external returns (uint256) {
+        return _collectRewards();
+    }
+
+    /// @dev Collect pending rewards
+    /// @return rewardAmount Amount of reward collected
+    function _collectRewards() internal returns (uint256 rewardAmount) {
```
DanielVF, line 72: Given that this should only be called externally (see other comment) let's just make this implementation the external one.

contracts/OGNRewardsSource.sol
```diff
@@ -0,0 +1,156 @@
...
+
+    /// @dev Set the rate of reward emission
+    /// @param _rewardsPerSecond Amount of OGN to distribute per second
+    function _setRewardsPerSecond(uint256 _rewardsPerSecond) internal {
+        if (_rewardsPerSecond > type(uint192).max) {
+            revert InvalidRewardRate();
+        }
+
+        // Collect any pending rewards at current rate
+        _collectRewards();
```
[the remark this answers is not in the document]
shahthepro: > The staking contract can only track rewards that happen when it calls the rewards source.

Good catch, will change the call and make it callable only by the `rewardsTarget` (which would be the Staking contract)

contracts/OGNRewardsSource.sol
```diff
@@ -0,0 +1,156 @@
...
+    error UnauthorizedCaller();
+    error RewardsTargetNotSet();
+    error InvalidRewardRate();
+
+    event StrategistUpdated(address _address);
+    event RewardsTargetChange(address target, address previousTarget);
+    event RewardsPerSecondChanged(uint256 newRPS, uint256 oldRPS);
+    event RewardCollected(uint256 amountCollected);
+
+    address public immutable ogn;
```
DanielVF, line 18: Let's change `ogn` to `asset` since it will be more generic.

contracts/OGNRewardsSource.sol
```diff
@@ -0,0 +1,156 @@
...
+    /// @dev Compute pending rewards since last collect
+    /// @return rewardAmount Amount of reward that'll be distributed if collected now
+    function previewRewards() external view returns (uint256) {
+        return _previewRewards(rewardConfig);
+    }
+
+    /// @dev Compute pending rewards since last collect
+    /// @param _rewardConfig RewardConfig
+    /// @return rewardAmount Amount of reward that'll be distributed if collected now
+    function _previewRewards(RewardConfig memory _rewardConfig) internal view returns (uint256) {
```
DanielVF, line 105: Is there any difference in gas usage from having an internal / and external _preview rewards, vs one public method?

contracts/OGNRewardsSource.sol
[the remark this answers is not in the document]
shahthepro: Have made the change

contracts/OGNRewardsSource.sol
shahthepro: Thought of that. But since the file name has "OGN" in it, I just went with `ogn` for this one as well. Will change it.

Also, IMO `rewardToken` might be a bit more apt than `asset`

shahthepro: For the filename, what do you think of `RewardsSourceV2`?

contracts/OGNRewardsSource.sol
shahthepro: Will run some tests and see if it makes any change

contracts/OGNRewardsSource.sol
DanielVF: I'm good with `rewardToken`.

Maybe `fixedRewardsSource`? It's more of a different kind of thing, than an improved version.

contracts/OGNRewardsSource.sol
shahthepro: Having a public method was slightly cheaper in terms of gas usage. Have changed the code

contracts/OGNRewardsSource.sol
```diff
@@ -0,0 +1,156 @@
...
+        if (_target == address(0)) {
+            revert RewardsTargetNotSet();
+        }
+
+        // Compute pending rewards
+        RewardConfig storage _config = rewardConfig;
+        rewardAmount = _previewRewards(_config);
+
+        // Update timestamp
+        _config.lastCollect = uint64(block.timestamp);
```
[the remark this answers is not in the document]
shahthepro: Tested this by Reading it as memory, updating the memory var and using that to write the global var. However it was slightly more expensive than the recent change (public `previewRewards` that doesn't accept any arg and reads it from memory)

contracts/FixedRateRewardsSource.sol
```diff
@@ -0,0 +1,140 @@
...
+    /// @param _rewardsTarget contract address that can collect rewards
+    function _setRewardsTarget(address _rewardsTarget) internal {
+        emit RewardsTargetChange(_rewardsTarget, rewardsTarget);
+        // Can be set to zero to disable
+        rewardsTarget = _rewardsTarget;
+    }
+
+    /// @dev Set the rate of reward emission
+    /// @param _rewardsPerSecond Amount of rewardToken to distribute per second
+    function setRewardsPerSecond(uint256 _rewardsPerSecond) external onlyGovernorOrStrategist {
```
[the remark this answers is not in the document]
shahthepro: Fair point, will do that

contracts/FixedRateRewardsSource.sol
```diff
@@ -0,0 +1,140 @@
...
+        rewardConfig.lastCollect = uint64(block.timestamp);
+
+        if (rewardAmount > 0) {
+            // Should not revert if there's no reward to transfer.
+
+            emit RewardCollected(rewardAmount);
+
+            // Intentionally skipping balance check to save some gas
+            // since `transfer` anyway would fail in case of low balance
+            IERC20(rewardToken).transfer(_target, rewardAmount);
```
pandadefi, line 84: In case the contract doesn't have enough tokens to transfer, this will revert.

contracts/FixedRateRewardsSource.sol
```diff
@@ -0,0 +1,140 @@
...
+    /// @param _rewardsPerSecond Amount of rewardToken to distribute per second
+    function _setRewardsPerSecond(uint256 _rewardsPerSecond) internal {
+        if (_rewardsPerSecond > type(uint192).max) {
+            revert InvalidRewardRate();
+        }
+
+        // Update storage
+        RewardConfig storage _config = rewardConfig;
+        emit RewardsPerSecondChanged(_rewardsPerSecond, _config.rewardsPerSecond);
+        _config.rewardsPerSecond = uint192(_rewardsPerSecond);
```
pandadefi, line 138: Updating the rewards per second should likely trigger the collection of rewards, transferring the accumulated rewards up to that point using the `_rewardsPerSecond` value before the update is applied.

pandadefi:
@shahthepro I have added my comments regarding the changes.

contracts/FixedRateRewardsSource.sol
shahthepro: Yep, that's expected. The staking contract will deal with error handling

contracts/FixedRateRewardsSource.sol
```diff
@@ -0,0 +1,140 @@
...
+    address public immutable rewardToken;
+
+    address public strategistAddr;
+
+    address public rewardsTarget;
+
+    struct RewardConfig {
+        // Inspired by (Copied from) `Dripper.Drip` struct.
+        uint64 lastCollect; // Overflows 262 billion years after the sun dies
+        uint192 rewardsPerSecond;
```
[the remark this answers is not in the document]
shahthepro: I don't foresee an end of rewards. @DanielVF thoughts?

contracts/FixedRateRewardsSource.sol
pandadefi: You could add a non `distributedRewards` variable to keep this for a later distribution.

contracts/FixedRateRewardsSource.sol
```diff
@@ -0,0 +1,136 @@
...
+            // since `transfer` anyway would fail in case of low balance
+            IERC20(rewardToken).transfer(_target, rewardAmount);
+        }
+    }
+
+    /// @dev Compute pending rewards since last collect
+    /// @return rewardAmount Amount of reward that'll be distributed if collected now
+    function previewRewards() public view returns (uint256) {
+        RewardConfig memory _config = rewardConfig;
+        return (block.timestamp - _config.lastCollect) * _config.rewardsPerSecond;
```
DanielVF, line 92: We should return zero here if `_config.lastCollect == 0`.

DanielVF, line 92: As much as I'm sad about one extra call's worth of gas. I think we probably do want to check the rewardToken's balance here and reduce down to it if it's not full. In this way if we stop feeding the contract extra funds, it will still continue to distribute the funds that it does get. In this way the rewards part fails to partially working even if there is no governance actions to change things.

contracts/FixedRateRewardsSource.sol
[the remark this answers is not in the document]
DanielVF: Would probably lean away from it. This is less of an immutable guarantee, and more that we are giving people tokens in a way that could change at any time.

contracts/FixedRateRewardsSource.sol
shahthepro: PR is here: https://github.com/OriginProtocol/ousd-governance/pull/413

DanielVF: @shahthepro Still need a fix on this one.

DanielVF (approved):
Approving FixedRateRewardsSource

tests/staking/Migrator.t.sol
```diff
@@ -0,0 +1,413 @@
...
+
+        vm.expectRevert();
+        (amount, end, points) = ognStaking.lockups(alice, 1);
+
+        // Should have removed OGV staked
+        for (uint256 i = 0; i < lockupIds.length; ++i) {
+            (amount, end, points) = ogvStaking.lockups(alice, lockupIds[i]);
+            assertEq(amount, 0, "Lockup still exists");
+            assertEq(end, 0, "Lockup still exists");
+            assertEq(points, 0, "Lockup still exists");
```
PraneshASP, line 408: good to add a different string as it will be easier to figure out which one fails, if one if them fail.

```solidity
  assertEq(amount, 0, "amount:Lockup still exists");
  assertEq(end, 0, "end:Lockup still exists");
  assertEq(points, 0, "points:Lockup still exists");
```

contracts/Migrator.sol
```diff
@@ -0,0 +1,228 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.10;
+
+import "OpenZeppelin/openzeppelin-contracts@4.6.0/contracts/token/ERC20/extensions/ERC20Burnable.sol";
+import "./Governable.sol";
+
+interface IStaking {
```
PraneshASP, line 7: can we move this interface to a separate (interfaces) folder?

tests/staking/Migrator.t.sol
```diff
@@ -0,0 +1,413 @@
...
+import "contracts/ExponentialStaking.sol";
+
+import "contracts/upgrades/MigratorProxy.sol";
+
+import "contracts/tests/MockOGN.sol";
+import "contracts/tests/MockRewardsSource.sol";
+import "contracts/tests/MockOGV.sol";
+import "contracts/tests/MockOGVStaking.sol";
+
+contract MigratorTest is Test {
```
PraneshASP, line 15: good to add tests for event data as well.

tests/staking/Migrator.t.sol
```diff
@@ -0,0 +1,413 @@
...
+
+        migrator = new Migrator(address(ogv), address(ogn), address(ogvStaking), address(ognStaking));
+        mProxy.initialize(address(migrator), governor, "");
+        migrator = Migrator(address(mProxy));
+
+        // Make sure contract has enough OGN for migration
+        ogn.mint(address(migrator), 10000000 ether);
+
+        // Users have enough OGV
+        ogv.mint(alice, 10000000 ether);
```
PraneshASP, line 55: `10000000 ether` seems redundant across the tests. maybe make it a constant?

tests/staking/Migrator.t.sol
```diff
@@ -0,0 +1,413 @@
...
+    function testMigrateStakes() public {
+        vm.startPrank(alice);
+        ogvStaking.mockStake(10000 ether, 365 days);
+        ogvStaking.mockStake(1000 ether, 20 days);
+
+        uint256[] memory lockupIds = new uint256[](2);
+        lockupIds[0] = 0;
+        lockupIds[1] = 1;
+
+        uint256 stakeAmount = (11000 ether * 0.09137 ether) / 1 ether;
```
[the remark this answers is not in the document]
DanielVF: I guess one advantages of having this number in the tests as well is that it confirms that the contract also has the right conversion rate set. Somewhat like double entry accounting.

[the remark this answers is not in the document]
shahthepro: Yeah, agree with both, I can probably move it to the top of the test file as constant (still separate from the hardcoded contract value)

PR #409, opened 2024-04-17
xOGN staking contract

DanielVF:

Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/contracts/ExponentialStaking.sol
+++ b/contracts/ExponentialStaking.sol
@@ -142,12 +142,12 @@ contract ExponentialStaking is ERC20Votes {
         } else {
             lockups[to].push(lockup);
             uint256 numLockups = lockups[to].length;
+            require(numLockups < uint256(type(int256).max), "Staking: Too many lockups");
+            lockupId = int256(numLockups - 1);
             // Delegate voting power to the receiver, if unregistered and first stake
             if (numLockups == 1 && delegates(to) == address(0)) {
                 _delegate(to, to);
             }
-            require(numLockups < uint256(type(int256).max), "Staking: Too many lockups");
-            lockupId = int256(numLockups - 1);
         }
         _mint(to, newPoints - oldPoints);
         emit Stake(to, uint256(lockupId), newAmount, newEnd, newPoints);
@@ -273,7 +273,7 @@ contract ExponentialStaking is ERC20Votes {
     ///         created so far (including expired & unstaked ones)
     /// @param user Address
     /// @return asset Number of lockups the user has had
-    function getLockupsCount(address user) external view returns (uint256) {
+    function lockupsCount(address user) external view returns (uint256) {
         return lockups[user].length;
     }
 }
--- a/contracts/ExponentialStaking.sol
+++ b/contracts/ExponentialStaking.sol
@@ -267,4 +267,12 @@ contract ExponentialStaking is ERC20Votes {
         (uint256 currentPoints,) = previewPoints(1e36, 0); // 1e36 saves a later multiplication
         return amount * ((currentPoints / fullPoints)) / 1e18;
     }
+
+    /// @notice Returns the total number of lockups the user has
+    ///         created so far (including expired & unstaked ones)
+    /// @param user Address
+    /// @return asset Number of lockups the user has had
+    function getLockupsCount(address user) external view returns (uint256) {
+        return lockups[user].length;
+    }
 }
--- a/contracts/ExponentialStaking.sol
+++ b/contracts/ExponentialStaking.sol
@@ -139,15 +139,14 @@ contract ExponentialStaking is ERC20Votes {
             require(newEnd > oldEnd, "Staking: New lockup must be longer");
             lockups[to][uint256(lockupId)] = lockup;
         } else {
-            lockupId = lockups[to].length;
-            require(lockupId < uint256(type(int256).max), "Staking: Too many lockups");
-
             lockups[to].push(lockup);
-
+            uint256 numLockups = lockups[to].length;
             // Delegate voting power to the receiver, if unregistered and first stake
-            if (lockupId == 0 && delegates(to) == address(0)) {
+            if (numLockups == 1 && delegates(to) == address(0)) {
                 _delegate(to, to);
             }
+            require(numLockups < uint256(type(int256).max), "Staking: Too many lockups");
+            lockupId = int256(numLockups - 1);
         }
         _mint(to, newPoints - oldPoints);
         emit Stake(to, uint256(lockupId), newAmount, newEnd, newPoints);
--- a/contracts/ExponentialStaking.sol
+++ b/contracts/ExponentialStaking.sol
@@ -1,0 +1,267 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.10;
+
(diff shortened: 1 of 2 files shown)


contracts/ExponentialStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+
+    /// @notice Preview the number of points that would be returned for the
+    /// given amount and duration.
+    ///
+    /// @param amount asset to be staked
+    /// @param duration number of seconds to stake for
+    /// @return points staking points that would be returned
+    /// @return end staking period end date
+    function previewPoints(uint256 amount, uint256 duration) public view returns (uint256, uint256) {
+        require(duration <= 1461 days, "Staking: Too long");
```
franckc, line 229: This should be maxStakeDuration vs 1461

contracts/ExponentialStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+        uint256 withdrawAmount = previewWithdraw(amount, end);
+        uint256 penalty = amount - withdrawAmount;
+
+        delete lockups[msg.sender][lockupId]; // Keeps empty in array, so indexes are stable
+        _burn(msg.sender, points);
+        if (penalty > 0) {
+            asset.transfer(address(rewardsSource), penalty);
+        }
+        asset.transfer(msg.sender, withdrawAmount);
+        emit Unstake(msg.sender, lockupId, withdrawAmount, end, points);
```
franckc, line 172: Wonder if we should add the penalty to the event to make it easier for:
1. analytics
2. off-chain calculation of how much total penalties occurred so that it can be factored in for updating the rate emission in the future (since penalty should be redistributed as extra rewards to other stakers).

contracts/ExponentialStaking.sol
DanielVF: So it should!

contracts/ExponentialStaking.sol
```diff
@@ -0,0 +1,269 @@
...
+    /// epoch.
+    ///
+    /// Any rewards previously earned will be paid out or rolled into the stake.
+    ///
+    /// @param amountIn asset to lockup in the stake
+    /// @param duration in seconds for the stake
+    /// @param to address to receive ownership of the stake
+    /// @param stakeRewards should pending user rewards be added to the stake
+    /// @param lockupId previous stake to extend / add funds to. -1 to create a new stake.
+    function stake(uint256 amountIn, uint256 duration, address to, bool stakeRewards, int256 lockupId) external {
```
shahthepro, line 91: I just realized that we also need an additional parameter to set the delegate (especially when it's gonna be called by the migrator). Let me pull this changeset to my branch and add that (or we could just let the user make two calls)

contracts/ExponentialStaking.sol
```diff
@@ -0,0 +1,269 @@
...
+        lockup.amount = uint128(newAmount); // max checked in require above
+        lockup.points = newPoints;
+
+        // Update or create lockup
+        if (lockupId != NEW_STAKE) {
+            require(newEnd > oldEnd, "Staking: New lockup must be longer");
+            lockups[to][uint256(lockupId)] = lockup;
+        } else {
+            lockups[to].push(lockup);
+            uint256 numLockups = lockups[to].length;
```
shahthepro, line 143: The stake event doesn't emit the `lockupId` in case of a new stake correctly. Probably change this line to `lockupId = lockups[to].length`, that should work

contracts/ExponentialStaking.sol
shahthepro: Pushed a minor fix for this: https://github.com/OriginProtocol/ousd-governance/pull/409/commits/3f51c2d70308499b4981396bbae19bb4920a6348

shahthepro: Nvm, I didn't see that it needed a typecast. Reverted that and have casted it properly. Also, fixed a few casing issues, it should work fine on CI now

contracts/ExponentialStaking.sol
```diff
@@ -0,0 +1,270 @@
...
+        require(newAmount <= type(uint128).max, "Staking: Too much");
+        (uint256 newPoints, uint256 newEnd) = previewPoints(newAmount, duration);
+        require(newPoints + totalSupply() <= type(uint192).max, "Staking: Max points exceeded");
+        lockup.end = uint128(newEnd);
+        lockup.amount = uint128(newAmount); // max checked in require above
+        lockup.points = newPoints;
+
+        // Update or create lockup
+        if (lockupId != NEW_STAKE) {
+            require(newEnd > oldEnd, "Staking: New lockup must be longer");
```
pandadefi, line 139: Is it mandatory to increase the duration of a lock if I want to increased the amount deposited?

contracts/ExponentialStaking.sol
```diff
@@ -0,0 +1,270 @@
...
+    ///
+    /// @param amount asset to be staked
+    /// @param duration number of seconds to stake for
+    /// @return points staking points that would be returned
+    /// @return end staking period end date
+    function previewPoints(uint256 amount, uint256 duration) public view returns (uint256, uint256) {
+        require(duration <= maxStakeDuration, "Staking: Too long");
+        uint256 start = block.timestamp > epoch ? block.timestamp : epoch;
+        uint256 end = start + duration;
+        uint256 endYearpoc = ((end - epoch) * 1e18) / 365 days;
```
pandadefi, line 235: Can you explain this part of the code?

DanielVF: There's basically a prestaking phase where all stakes are equal until staking officially begins.

So the start date of a stake made during prestaking is the official start of all staking (epoch)

The end date is calculated based on the duration from this start date.

To calculate the multiplier that the staking points get, we need to know many years the end date of the stake is from the start of staking.

contracts/ExponentialStaking.sol
DanielVF: It's probably fine either way, but I'll change it to allow increasing only the duration.

contracts/ExponentialStaking.sol
pandadefi: What I am not sure to understand is since `maxStakeDuration` is 365 days, isn't everyone going to have the same year?

DanielVF: Let's say:

Staking start day is June 4th, 2024

Current day is May 5th. User stakes for one year. The start date gets pushed up to June 4th, 2024. The end date is approximately June 5th, 2024. 

Then end date - staking start day is 1.0 years, which means that 1.4**1.0 = the user will get 1.4 staking points per OGN staked.

DanielVF: The "years" are 1e18 fixed point decimals, and so users are getting a different "year" for every different end timestamp.

contracts/ExponentialStaking.sol
DanielVF: Changed here https://github.com/OriginProtocol/ousd-governance/pull/409/commits/37843333f8a98e0feeb0fd6b6ebda6145a243fe0

contracts/ExponentialStaking.sol
pandadefi: ok understood.

shahthepro (approved):
LGTM

PR #410, opened 2024-04-21
Add Migrator

shahthepro:

Contract change checklist:
  - [ ] Unit tests pass


contracts/Migrator.sol
```diff
@@ -0,0 +1,226 @@
...
+    /**
+     * @notice Solvency Checks
+     *
+     * This ensures that the contract never transfers more than
+     * desired OGN amount in any case. This takes a balance diff
+     * of OGV and OGN and makes sure that the difference adds up.
+     * However, it doesn't revert if the difference is in favour
+     * of the contract (i.e. less OGN sent than expected).
+     */
+    modifier netBalanceCheck() {
```
DanielVF, line 111: Let's move a modifier to the front of the contract. As we discussed, let's make this an isSolvent check that ensures that all future migration can happen.

contracts/Migrator.sol
```diff
@@ -0,0 +1,226 @@
...
+    function start() external onlyGovernor {
+        if (endTime != 0) {
+            revert MigrationAlreadyStarted();
+        }
+
+        // Max approve
+        ogn.approve(address(ognStaking), type(uint256).max);
+
+        endTime = block.timestamp + 365 days;
+    }
```
DanielVF, line 60: Let's do our isSolvent check here as well, making sure that as soon as this is called, the migration is good to go.

contracts/Migrator.sol
```diff
@@ -0,0 +1,226 @@
...
+
+        uint256 ognBalance = ogn.balanceOf(address(this));
+        if (ognBalance > 0) {
+            // OGN doesn't allow burning of tokens. Has `onlyOwner`
+            // modifier on `burn` and `burnFrom` methods. Also,
+            // `transfer` has a address(0) check. So, this transfers
+            // everything to address(1). The `owner` multisig of
+            // OGN token can call `burnFrom(address(1))` later.abi
+
+            ogn.transfer(address(1), ognBalance);
```
DanielVF, line 90: Let's do `0x000000000000000000000000000000000000dead` instead of `0x1` Makes things clearer that this is a burn.

contracts/Migrator.sol
```diff
@@ -0,0 +1,226 @@
...
+     */
+    function migrateAll() external netBalanceCheck returns (uint256 ognReceived) {
+        return _migrate(ogv.balanceOf(msg.sender), msg.sender);
+    }
+
+    /**
+     * @notice Migrates caller's OGV to OGN and sends it to the `receiver`
+     * @return ognReceived OGN Received
+     */
+    function _migrate(uint256 ogvAmount, address receiver) internal returns (uint256 ognReceived) {
```
DanielVF, line 149: Let's move this internal _migrate to the end of the two migrate methods. This way it's after the two publicly facing migrate methods.

contracts/OgvStaking.sol
```diff
@@ -222,45 +246,40 @@ contract OgvStaking is ERC20Votes {
...
         rewardDebtPerShare[user] = accRewardPerShare;
+
         if (netRewards == 0) {
-            return;
+            return 0;
         }
+
         ogv.transfer(user, netRewards);
         emit Reward(user, netRewards);
     }
```
DanielVF, line 285: Maybe a comment on _delegate on why we are keeping it in the code even though it's not used.

contracts/OgvStaking.sol
```diff
@@ -222,45 +246,40 @@ contract OgvStaking is ERC20Votes {
...
-            try rewardsSource.collectRewards() {}
-            catch {
-                // Governance staking should continue, even if rewards fail
-            }
-            uint256 collected = ogv.balanceOf(address(this)) - preBalance;
-            accRewardPerShare += (collected * 1e12) / supply;
+    /// @param netRewards Net reward collected for user
+    function _collectRewards(address user) internal returns (uint256 netRewards) {
+        if (totalSupply() == 0) {
+            return 0; // No one has any points to even get rewards
```
DanielVF, line 271: I guess because points are both collected before unstaking, and new sakes are disabled, the system can never have zero points and still have undistributed rewards, so this is correct. Does look the wrong change at first glance though because the previous 0 > guard is for on collecting points into the contract's accounting, not on distributing points which the new if statement is guarding.

contracts/OgvStaking.sol
```diff
@@ -212,7 +235,8 @@ contract OgvStaking is ERC20Votes {
     // 3. Reward functions
 
     /// @notice Collect all earned OGV rewards.
-    function collectRewards() external {
+    /// @return rewardCollected OGV reward amount collected
+    function collectRewards() external returns (uint256 rewardCollected) {
         _collectRewards(msg.sender);
```
DanielVF, line 240: Does this need a return statement?

contracts/OgvStaking.sol
```diff
@@ -172,24 +212,7 @@ contract OgvStaking is ERC20Votes {
...
-        if (!hasDelegationSet[msg.sender] && delegates(msg.sender) == address(0)) {
-            // Delegate voting power to the receiver, if unregistered
-            _delegate(msg.sender, msg.sender);
-        }
-        emit Unstake(msg.sender, lockupId, oldAmount, oldEnd, oldPoints);
-        emit Stake(msg.sender, lockupId, oldAmount, newEnd, newPoints);
+        revert StakingDisabled();
     }
 
     /// @notice Preview the number of points that would be returned for the
```
DanielVF, line 218: previewPoints should revert with stakingDisabled.

contracts/OgvStaking.sol
```diff
@@ -108,54 +125,77 @@ contract OgvStaking is ERC20Votes {
...
+            unstakedAmount += amount;
+
+            // Make sure it isn't unstaked already
+            if (end == 0) {
+                revert AlreadyUnstaked(lockupId);
+            }
+
+            delete lockups[staker][lockupId]; // Keeps empty in array, so indexes are stable
+
+            _burn(staker, points);
```
DanielVF, line 195: Since lots of users have lots of stakes, let's total up the points to be burned and the OGV to be transferred, and do them all at the end. Will save a bunch of gas.

contracts/OgvStaking.sol
```diff
@@ -108,54 +125,77 @@ contract OgvStaking is ERC20Votes {
...
-
-        if (!hasDelegationSet[to] && delegates(to) == address(0)) {
-            // Delegate voting power to the receiver, if unregistered
-            _delegate(to, to);
-        }
+    /// @notice Collect staked OGV for a lockup and any earned rewards.
+    /// @param lockupId the id of the lockup to unstake
+    /// @return unstakedAmount OGV amount unstaked
+    /// @return rewardCollected OGV reward amount collected
+    function unstake(uint256 lockupId) external returns (uint256 unstakedAmount, uint256 rewardCollected) {
```
DanielVF, line 135: Big thumbs up on driving all the external methods through a single internal method. Much cleaner code.

contracts/OgvStaking.sol
shahthepro: Good point, will make that change

contracts/OgvStaking.sol
shahthepro: It's still being used in the contracts it inherits from though. However there's no reason why someone would wanna delegate after this goes live

contracts/OgvStaking.sol
shahthepro: Done

contracts/OgvStaking.sol
shahthepro: Done

contracts/OgvStaking.sol
shahthepro: Yep, fixed it

contracts/Migrator.sol
shahthepro: This is done as well

contracts/Migrator.sol
shahthepro: Changed it to `address(0xdead)`

contracts/Migrator.sol
shahthepro: Have added that here and for `transferExcessTokens()` method too

contracts/Migrator.sol
shahthepro: Have made that change. Now the contract only checks if it can fund the migration of OGV still in circulation (or staked)

contracts/Migrator.sol
```diff
@@ -87,112 +108,73 @@ contract Migrator is Governable {
...
         bool migrateRewards,
+        uint256 newStakeAmount,
         uint256 newStakeDuration
-    ) external netBalanceCheck {
+    ) external isSolvent {
         if (!isMigrationActive()) {
             revert MigrationIsInactive();
         }
 
+        if (newStakeAmount == 0) {
```
DanielVF, line 170: It's fine for new stake amount to be zero - we just don't do a stake in that case.

contracts/Migrator.sol
```diff
@@ -0,0 +1,239 @@
...
+    function _migrate(uint256 ogvAmount, address receiver) internal returns (uint256 ognReceived) {
+        if (!isMigrationActive()) {
+            revert MigrationIsInactive();
+        }
+
+        ognReceived = (ogvAmount * CONVERSION_RATE) / 1 ether;
+
+        emit TokenExchanged(ogvAmount, ognReceived);
+
+        ogv.transferFrom(msg.sender, address(this), ogvAmount);
```
pandadefi, line 228: With the user allowance, you could use `burnFrom` reducing slightly the gas usage.

shahthepro: The actual spec was to not burn any tokens until the contract is decommissioned

contracts/Migrator.sol
```diff
@@ -0,0 +1,239 @@
...
+            // Burn all OGV
+            ogv.burn(ogvBalance);
+        }
+
+        uint256 ognBalance = ogn.balanceOf(address(this));
+        if (ognBalance > 0) {
+            // OGN doesn't allow burning of tokens. Has `onlyOwner`
+            // modifier on `burn` and `burnFrom` methods. Also,
+            // `transfer` has a address(0) check. So, this transfers
+            // everything to address(1). The `owner` multisig of
```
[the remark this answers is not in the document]
shahthepro: Fixed

contracts/Migrator.sol
DanielVF: Checking against totalSupply is beautiful way to do this. Way better than what I was thinking.

contracts/Migrator.sol
DanielVF: Given that we've changed to running the checks off the ogv totalSupply instead of balances, we could switch to just burning ogv as it comes in. I would be glad to do this.

DanielVF: This would also get rid of the need to account for ogv held by the contract in the solvency check.

contracts/Migrator.sol
```diff
@@ -0,0 +1,239 @@
...
+
+        emit LockupsMigrated(msg.sender, lockupIds, newStakeAmount, newStakeDuration);
+    }
+
+    /**
+     * @notice Migrates caller's OGV to OGN and sends it to the `receiver`
+     * @return ognReceived OGN Received
+     */
+    function _migrate(uint256 ogvAmount, address receiver) internal returns (uint256 ognReceived) {
+        if (!isMigrationActive()) {
```
DanielVF, line 220: We check migration active both in an external method and again in this internal method. This means that a migrate with lockups checks this twice. Usually it's better to have a check happen purely at one layer, either internally only, of externally only.

However, I think we would be okay to drop the isMigrationActive checks from the migration methods. This would mean that migrations could continue until we decommission, but that should be fine. Will make for less gas during migrations.

We decommission, the OGN leaving the contract will stop migrations.

It's fine if we allow coin migrations before the official start. staking migrations won't work, of course until start has been called.

shahthepro: That makes sense. Have made the change. However do note that people can start using the Migrator contract as soon as it's deployed and funded.

So, let's say we send 10M OGN to the contract for migrations but we only need 9M or so (and haven't called `transferExcessTokens` yet), There's a chance that the `isSolvent` won't catch if for some reason people get more OGN than expected for that additional 1M OGN

DanielVF: Good points.

We'll just have to keep testing amounts in it until we are ready to go, and perhaps our go live action will pull any excess from it just before/after the start().

DanielVF:
# Code Review - veOGV


## Requirements

- Instant unstakes
- Migrator controlled unstakes to the user
- All staking disabled

#### Authentication

- [x] Never use tx.origin
- [x] Every external/public function is supposed to be externally accessible
- [x] Every external/public function has the correct authentication

#### Ethereum

#### Cryptographic code

_no crypto_

#### Gas problems

- [x] Contracts with for loops must have either:
    - [ ] A way to remove items
    - [x] Size can be changed by user
    - [ ] Can be upgraded to get unstuck
    - [ ] Size can only controlled by admins
- [x] Contracts with for loops must not allow end users to add unlimited items to a loop that is used by others or admins.

#### Black magic

- [x] Does not contain `selfdestruct`
- [x] Does not use `delegatecall` outside of proxying. _If an implementation contract were to call delegatecall under attacker control, it could call selfdestruct the implementation contract, leading to calls through the proxy silently succeeding, even though they were failing._
- [x] Address.isContract should be treated as if could return anything at any time, because that's reality.

#### Overflow

- [x] Code is solidity version >= 0.8.0
- [x] All for loops use uint256

#### Proxy

- [x] No storage variable initialized at definition when contract used as a proxy implementation.

#### Events
- [x] All state changing functions emit events

#### Rounding
- [x] Contract rounds in the protocols favor
- [x] Contract does not have bugs from loosing rounding precision
- [x] Code correctly multiplies before division
- [x] Contract does not have bugs from zero or near zero amounts

#### Dependencies
- [x] Review any new contract dependencies thoroughly (e.g. OpenZeppelin imports) when new dependencies are added or version of dependencies changes.
- [x] If OpenZeppelin ACL roles are use review & enumerate all of them.
- [x] Check OpenZeppelin [security vulnerabilities](https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories) and see if any apply to current PR considering the version of OpenZeppelin contract used.

#### External calls

- [x] Contract addresses passed in are validated
- [x] No unsafe external calls
- [x] Reentrancy
		- [ ] guards on all state changing functions
		- [x] Only interact with trusted contracts
- [x] No malicious behaviors
- [x] Low level call() must require success.
- [x] No slippage attacks (we need to validate expected tokens received)
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent
  - [ ] If oracle can be bent, it won't hurt us.
- [x] Do not call balanceOf for external contracts to determine what they will do when they use internal accounting

#### Deploy

_No deploy code written_

When deploy code is written, a collect call should be made just before this upgrade.

#### Logic

Logic appears correct.

#### Internal State

- [x] accRewardsPerShare no longer changes
- [x] No new lockups can be pushed 

#### Attack

This contract holds 3.7 billion OGV, which will soon be convertible to OGN - 40 something million at today's prices.

An attack would attempt to gain access to another user's tokens,  get more tokens than owed, or double withdraw.

- [x] All unstakes transfer OGV to the stake owner
- [x] The amount calculations to send to the user are a simple sum of staked amounts being cleared
- [x] Stakes are disabled in storage as soon as amounts are added to total
- [x] Duplicate stake ids in a removal should fail 

#### Flavor

The code is slightly awkward because of the tension between keeping backwards compatibility versus most functionality no longer being needed. However there is not any anticipated future work on this contract, and it's okay as is.

DanielVF:
Noting that we should do some fork testing of this during the audit.

DanielVF (approved):

PR #418, opened 2024-05-14
Foundry based deploy script

DanielVF:


DanielVF:
Starting test:

`IS_FORK=true forge script --target-contract OgnOgvMigrationScript --rpc-url="URL_HERE" -vvvv script/deploy/mainnet/010_OgnOgvMigrationScript.sol`

DanielVF:
Long term goal is to have the base script expose a registry for deployed contracts.

This way when you make a fork test, you run the script, then use the registry to get to the latest deployed contract addresses.

Also, the idea would be to return a governance object that knows how to both execute its commands in a fork test, but also return something useful for submitting to governance.

We'll have to support both governance proposals AND multisig to direct to timelock for these migrations. Probably our highest priority though is getting some good fork tests to run after these deploys run.

script/deploy/mainnet/BaseMainnetScript.sol
```diff
@@ -0,0 +1,48 @@
...
+        if (block.chainid != 1) {
+            revert("Not Mainnet");
+        }
+        // Will not execute script if after this block number
+        if (block.number > deployBlockNum) {
+            // console.log("Current block %s, script block %s", block.number, deployBlockNum);
+            return;
+        }
+
+        isForked = vm.envOr("IS_FORK", false);
```
shahthepro, line 23: `vm.isContext(VmSafe.ForgeContext.ScriptDryRun)` might be a much cleaner way than environment variables for this one. Will make this change

shahthepro (approved):

PR #419, opened 2024-05-15
Add fork test tooling

shahthepro:

Contract change checklist:
  - [ ] Unit tests pass


script/deploy/mainnet/012_xOGNGovernanceScript.sol
```diff
@@ -0,0 +1,99 @@
...
+        for (uint256 i = 0; i < prop.actions.length; i++) {
+            GovFiveAction memory action = prop.actions[i];
+            bytes memory sig = abi.encodePacked(bytes4(keccak256(bytes(action.fullsig))));
+            vm.prank(Addresses.TIMELOCK);
+            action.receiver.call(abi.encodePacked(sig, action.data));
+        }
+    }
+}
+
+contract XOGNGovernanceScript is BaseMainnetScript {
```
DanielVF, line 64: In the real world this will be the multisig calling the timelock. However this is a lower priority. First we need to make sure that main switch over has enough tests on it.

script/deploy/mainnet/010_xOGNSetupScript.sol
```diff
@@ -0,0 +1,65 @@
...
+        ExponentialStaking xognImpl =
+            new ExponentialStaking(Addresses.OGN, ognEpoch, ognMinStaking, address(ognRewardsSourceProxy));
+        _recordDeploy("XOGN_IMPL", address(xognImpl));
+
+        console.log("- xOgnProxy init");
+        xOgnProxy.initialize(address(xognImpl), Addresses.TIMELOCK, "");
+
+        //
+        // 3. Rewards implimentation and init
+        uint256 rewardsPerSecond = 0; //TODO: Decide on the params
```
DanielVF, line 54: This will be 300,000 per day.  

Probably easiest to write it for clarity as the per day rate divided by the number of seconds in a day.

contracts/FixedRateRewardsSource.sol
```diff
@@ -90,10 +90,6 @@ contract FixedRateRewardsSource is Governable, Initializable {
     function previewRewards() public view returns (uint256 rewardAmount) {
         RewardConfig memory _config = rewardConfig;
 
-        if (_config.lastCollect == 0) {
```
sparrowDom, line 93: Currently deploy script 10 deploys the contract and deploy script 12 sets the rewards per second. I would leave this check in since it is a sanity check and not very gas expensive one at that. 

With the current code of the contract, the contract needs to be deployed and funded in the correct order, or it will not function properly. Since one can: 
- run deploy script 10 that leaves the rewardsPerSecond to 0
- fund the contract
- call collectRewards and the function will probably revert, since previewRewards will return a too big of a number.

I think the correct functioning of the contract should be assured even if the deploy is not configured perfectly as expected. For that reason I would: 
- re-introduce the above removed lines of code and add comments why it is imperative that they are present
- remove the potentially confusing `_rewardsPerSecond` from the initialize function. Any deployment of this contract will suffer the same issues on initialisation of rewards source and start of emission. 
- add additional comments explicitly stating that the contract is basically paused until the `setRewardsPerSecond` isn't called with a non 0 value.
- currently it is also possible that the rewards source is already running (funded and non 0 rewardsPerSecond) and governor calls the `setRewardsPerSecond` with a 0 value. This would in effect pause the contract and remove any rewards accrued since the last collect. Is that intentional? If not change the `_setRewardsPerSecond` to:
  - either make it impossible to call the function with the 0 parameter 
  - or collect the rewards and send it to a configured target. And pause the rewards source by setting the rewards to 0

sparrowDom: Thanks for correcting me on first 2 points; you're right they are not an issue.

contracts/FixedRateRewardsSource.sol
```diff
@@ -139,6 +135,11 @@ contract FixedRateRewardsSource is Governable, Initializable {
         // Update storage
         RewardConfig storage _config = rewardConfig;
         emit RewardsPerSecondChanged(_rewardsPerSecond, _config.rewardsPerSecond);
+        if (_config.rewardsPerSecond == 0) {
+            // When changing rate from zero to non-zero,
+            // Update lastCollect timestamp as well
```
sparrowDom, line 140: I only have 1 more nitpick: 
I would expand this comment to
```
/* This contract code allows for contract deployment & initialization and then the contract can be live for quite
 * some time before it is funded and `_rewardsPerSecond` are set to non 0 value. In that case the vesting period 
 * from contract initialization until now would be taken into account instead of the time since the contract has been
 * "activated" by setting the `setRewardsPerSecond`. To mitigate the issue we update the `_config.lastCollect`
 * to current time. 
 */
```

contracts/FixedRateRewardsSource.sol
DanielVF: I'm good with a change to rewards rate potentially changing how much users receive for the past, as well as the future. We'll probably usually couple it with a xOGN collect in most changes though.

Having a "pause" feature that stops rewards could be a useful thing.

script/deploy/mainnet/011_OgnOgvMigrationScript.sol
```diff
@@ -0,0 +1,155 @@
...
+        Migrator migratorImpl = new Migrator(Addresses.OGV, Addresses.OGN, Addresses.VEOGV, xOgnProxy);
+        _recordDeploy("MIGRATOR_IMPL", address(migratorImpl));
+
+        console.log("- Migrator init");
+        migratorProxy.initialize(address(migratorImpl), Addresses.TIMELOCK, "");
+
+        _buildGnosisTx();
+    }
+
+    function _buildGnosisTx() internal {
```
DanielVF, line 60: I'd rather call this something like `_buildGovernance`. Most of the time gnosis will only indirectly be involved, and we will be making governance proposals. Even outside this we'll talking to the timelock.

script/deploy/mainnet/011_OgnOgvMigrationScript.sol
```diff
@@ -0,0 +1,155 @@
...
+            "revokeRole(bytes32,address)",
+            abi.encode(timelock.CANCELLER_ROLE(), Addresses.GOVERNOR_FIVE)
+        );
+        govFive.action(
+            Addresses.TIMELOCK,
+            "revokeRole(bytes32,address)",
+            abi.encode(timelock.EXECUTOR_ROLE(), Addresses.GOVERNOR_FIVE)
+        );
+
+        govFive.action(Addresses.OUSD_BUYBACK, "upgradeTo(address)", abi.encode(Addresses.OUSD_BUYBACK_IMPL)); // Todo, use latest deployed address
```
DanielVF, line 89: Are these TODO's still TODO's?

script/deploy/mainnet/011_OgnOgvMigrationScript.sol
```diff
@@ -0,0 +1,155 @@
...
+        IMintableERC20 ogv = IMintableERC20(Addresses.OGV);
+        // Mint additional OGN, will get returned after starting migration
+        uint256 ognToMint = ((ogv.totalSupply() * 0.09137 ether) / 1 ether) + 1_000_000 ether;
+
+        address[] memory targets = new address[](1);
+        string[] memory sigs = new string[](1);
+        bytes[] memory calldatas = new bytes[](1);
+
+        // OGN Gov 1: Mint OGN
+        targets[0] = Addresses.OGN;
```
DanielVF, line 104: TODO: This will change when the ownership on OGN changes to the DeFi timelock

script/deploy/DeployManager.sol
```diff
@@ -0,0 +1,146 @@
...
+        for (uint256 i = 0; i < records.length; ++i) {
+            string memory name = records[i].name;
+            address addr = records[i].addr;
+
+            console.log(string(abi.encodePacked("> Recorded Deploy of ", name, " at")), addr);
+            networkDeployments = vm.serializeAddress(contractsKey, name, addr);
+            deployedContracts[name] = addr;
+        }
+
+        vm.writeJson(networkDeployments, deploymentsFilePath, contractsKey);
```
DanielVF, line 129: I'm still getting a lot of random failure on these writeJson's (about 2/3 of the 50% of tests fail, about 1/6 of the time, all tests succeed)

shahthepro: Yeah. It's mostly about writing twice to the same file back to back. Thats why I added a small delay. Will work on a clean fix to do a single write instead

contracts/FixedRateRewardsSource.sol
shahthepro: This is done

PR #427, opened 2024-05-23
Comments on clearing last collect

DanielVF:
During the audit we made the following two changes:

Removed from `FixedRateRewardsSource.sol` `initialize`:

```solidity
 _setRewardsPerSecond(_rewardsPerSecond); // removed
 ```
 
 Added to `FixedRateRewardsSource.sol` `_setRewardsPerSecond`:
 
 ```solidity
 if (_config.rewardsPerSecond == 0) {
    _config.lastCollect = uint64(block.timestamp);
}
```

These make it so that when rewards are enabled on the contract, the `lastCollect` is reset to the current time, and thus no rewards will be paid out for the period when the rewards were set to zero.

In other case, as before, the contract will pay out pending rewards from the lastCollect at the newly set rate. Calling collectRewards before setting the rate allows precise reward payments for past rewards.

This PR just updates the comments on this change.


contracts/FixedRateRewardsSource.sol
```diff
@@ -130,12 +130,13 @@ contract FixedRateRewardsSource is Governable, Initializable {
...
-             * some time before it is funded and `_rewardsPerSecond` are set to non 0 value. In that case the vesting period 
-             * from contract initialization until now would be taken into account instead of the time since the contract has been
-             * "activated" by setting the `setRewardsPerSecond`. To mitigate the issue we update the `_config.lastCollect`
-             * to current time.
-             */
+            // If rewardsPerSecond is zero, then there are no past rewards accrued,
+            // and it is safe to reset the last rewards time to the current time.
+            // We do this so that when enabling rewards after a period of not having rewards,
+            // we don’t pay out rewards for time before the rewards are enabled.
+            //
```
naddison36: the linter doesn't like the space after the `//`. I've pushed a change to fix

naddison36 (approved):

contracts/FixedRateRewardsSource.sol
DanielVF: 🙏

shahthepro (approved):

PR #429, opened 2024-05-25
Do Governance Proposals with Signatures

DanielVF:
Our governance contract has two different `propose` methods. One tracks the plain text signatures of called methods, and the other does not. We want to switch to using the signature based method for governance, since this makes verification easier, and our tooling is already built to use it.

Before:

`propose(address[],uint256[],bytes[],string)`

After:

`propose(address[],uint256[],string[],bytes[],string)`


contracts/utils/GovProposalHelper.sol
emagici, line 1: need to add the description for the functions and parameters

PR #435, opened 2024-05-30
Migrator changes

shahthepro:
- Adds support for migrating just the rewards (without any lockup/balance)
- Adds support for calling `migrate()` without `lockupIds`

This also removes the need for the `MigrationZapper` contract. The frontend can always call a single method for all purposes

--------

Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/contracts/Migrator.sol
+++ b/contracts/Migrator.sol
@@ -25,7 +25,6 @@ contract Migrator is Governable {
 
     error MigrationAlreadyStarted();
     error ContractInsolvent(uint256 expectedOGN, uint256 availableOGN);
-    error LockupIdsRequired();
     error InvalidStakeAmount();
 
     constructor(address _ogv, address _ogn, address _ogvStaking, address _ognStaking) {
@@ -131,12 +130,12 @@ contract Migrator is Governable {
             // Unstake if there are any lockups
             (uint256 ogvAmountUnlocked, uint256 rewardsCollected) = ogvStaking.unstakeFrom(msg.sender, lockupIds);
 
+            ogvAmountFromWallet += ogvAmountUnlocked;
+
             if (migrateRewards) {
                 // Include rewards if needed
                 ogvAmountFromWallet += rewardsCollected;
             }
-
-            ogvAmountFromWallet += ogvAmountUnlocked;
         } else if (migrateRewards) {
             uint256 rewardsCollected = ogvStaking.collectRewardsFrom(msg.sender);
             ogvAmountFromWallet += rewardsCollected;
--- a/contracts/Migrator.sol
+++ b/contracts/Migrator.sol
@@ -127,19 +127,20 @@ contract Migrator is Governable {
         uint256 newStakeAmount,
         uint256 newStakeDuration
     ) external isSolvent {
-        if (lockupIds.length == 0) {
-            revert LockupIdsRequired();
-        }
+        if (lockupIds.length > 0) {
+            // Unstake if there are any lockups
+            (uint256 ogvAmountUnlocked, uint256 rewardsCollected) = ogvStaking.unstakeFrom(msg.sender, lockupIds);
 
-        // Unstake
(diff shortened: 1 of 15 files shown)


contracts/Migrator.sol
```diff
@@ -127,20 +127,21 @@ contract Migrator is Governable {
...
-
-        if (migrateRewards) {
-            // Include rewards if needed
+        if (lockupIds.length > 0) {
+            // Unstake if there are any lockups
+            (uint256 ogvAmountUnlocked, uint256 rewardsCollected) = ogvStaking.unstakeFrom(msg.sender, lockupIds);
+
+            if (migrateRewards) {
+                // Include rewards if needed
+                ogvAmountFromWallet += rewardsCollected;
```
pandadefi, line 136: Is the approval on the frontend adding extra?

shahthepro: Yep. It already happens when someone tries to unstake lockups with rewards. This changeset just enables users to claim rewards and migrate them without unstaking any lockups

contracts/utils/GovFive.sol
```diff
@@ -6,6 +6,7 @@ import {Vm} from "forge-std/Vm.sol";
 import {Addresses} from "contracts/utils/Addresses.sol";
 import "forge-std/console.sol";
 
+import {TimelockController} from "OpenZeppelin/openzeppelin-contracts@4.6.0/contracts/governance/TimelockController.sol";
 import "OpenZeppelin/openzeppelin-contracts@4.6.0/contracts/utils/Strings.sol";
 
 library GovFive {
```
DanielVF, line 12: I think we should spin timelock governance control out out into a seperate file from the governor bravo style governance submission. We'll be moving back to the governor bravo style in the future.

shahthepro: We already have a `GovProposalHelper` for Bravo style governance. This one only deal with multisig governance

DanielVF: (an idea is that the difference governance methods can be pluggable, this way we can use this code on different chains with difference governance systems.)

contracts/Migrator.sol
```diff
@@ -127,20 +127,21 @@ contract Migrator is Governable {
         uint256 newStakeAmount,
         uint256 newStakeDuration
     ) external isSolvent {
-        if (lockupIds.length == 0) {
-            revert LockupIdsRequired();
```
DanielVF, line 131: We can remove the `LockupIdsRequired` error message.

contracts/Migrator.sol
```diff
@@ -127,20 +127,21 @@ contract Migrator is Governable {
...
+        if (lockupIds.length > 0) {
+            // Unstake if there are any lockups
+            (uint256 ogvAmountUnlocked, uint256 rewardsCollected) = ogvStaking.unstakeFrom(msg.sender, lockupIds);
+
+            if (migrateRewards) {
+                // Include rewards if needed
+                ogvAmountFromWallet += rewardsCollected;
+            }
+
+            ogvAmountFromWallet += ogvAmountUnlocked;
```
DanielVF, line 139: This would be clearer if the `ogvAmountFromWallet += ogvAmountUnlocked;` is moved to be right under the `unstakeFrom`, since that's where the `ogvAmountUnlocked` value comes from. I like having variables that have a short lifetime in the code.

shahthepro: Done

contracts/Migrator.sol
shahthepro: I thought I removed it, not sure how it slipped, have removed it now

contracts/utils/GovFive.sol
shahthepro: It should be possible to do that but I'll put that one up as a different PR

DanielVF: Maybe I'm just confused by the name `GovFive` which is the name of our OGV Governance contract. Maybe all that needs to change here is the name of this contract then.

DanielVF (approved):

